PR #1, opened 2018-06-01
add tcp, tcp-tls features + refactor of client, server

jkralik:


client.go
```diff
@@ -0,0 +1,295 @@
...
+	case *net.TCPConn, *tls.Conn:
+		resp, _, err := PullTcp(m)
+		if err != nil {
+			return nil, err
+		}
+		return resp, nil
+	}
+	// UDP connection
+	resp, err := ParseDgramMessage(m)
+	if err != nil {
```
michalpristas: you can omit check and return resp, err

michalpristas:
few pointers.
- read this: https://medium.com/@benbjohnson/standard-package-layout-7cdbc8391fc1
- try to use more granular files. they are long and not readable, example client.go, it contains definition of client, connection and some statics. crate client.go and conn.go 
- stick with interfaces, then create struct implementing interace. keep interfaces small. combine them to larger and more powerfull ones

this might be of use as well: https://www.youtube.com/watch?v=MzTcsI6tn-0

i can look into specifics later

README.md
```diff
@@ -0,0 +1,180 @@
...
+		// for tcp
+		// co, err := coap.Dial("tcp", "localhost:5688")
+		
+		// for tcp-tls
+		// co, err := coap.DialWithTLS("localhost:5688", &tls.Config{InsecureSkipVerify: true})
+
+		if err != nil {
+			log.Fatalf("Error dialing: %v", err)
+		}
+		defer co.Close()
```
michalpristas: close the connection right away, defer for close might lead to running out of sockets.
also co.Close() error is ignored

client.go
```diff
@@ -0,0 +1,295 @@
...
+			continue
+		} else if err != nil {
+			return nil, err
+		}
+		break
+	}
+	switch co.Conn.(type) {
+	case *net.TCPConn, *tls.Conn:
+		resp, _, err := PullTcp(m)
+		if err != nil {
```
michalpristas: same here return resp, err. turn on golint

client.go
```diff
@@ -0,0 +1,295 @@
...
+func (c *Client) getTimeoutForRequest(timeout time.Duration) time.Duration {
+	var requestTimeout time.Duration
+	if c.Timeout != 0 {
+		requestTimeout = c.Timeout
+	} else {
+		requestTimeout = timeout
+	}
+	// net.Dialer.Timeout has priority if smaller than the timeouts computed so
+	// far
+	if c.Dialer != nil && c.Dialer.Timeout != 0 {
```
michalpristas: readability: single if?

client.go
```diff
@@ -0,0 +1,295 @@
...
+		}
+	}
+	return requestTimeout
+}
+
+// Dial connects to the address on the named network.
+func Dial(network, address string) (conn *Conn, err error) {
+	client := Client{Net: network}
+	conn, err = client.Dial(address)
+	if err != nil {
```
michalpristas: return conn, err

michalpristas: even better, return client.DIal()

client.go
```diff
@@ -0,0 +1,295 @@
...
+		return nil, err
+	}
+	return conn, nil
+}
+
+// DialTimeout acts like Dial but takes a timeout.
+func DialTimeout(network, address string, timeout time.Duration) (conn *Conn, err error) {
+	client := Client{Net: network, Dialer: &net.Dialer{Timeout: timeout}}
+	conn, err = client.Dial(address)
```
michalpristas: same

client.go
```diff
@@ -0,0 +1,295 @@
...
+	return conn, nil
+}
+
+// DialWithTLS connects to the address on the named network with TLS.
+func DialWithTLS(network, address string, tlsConfig *tls.Config) (conn *Conn, err error) {
+	if !strings.HasSuffix(network, "-tls") {
+		network += "-tls"
+	}
+	client := Client{Net: network, TLSConfig: tlsConfig}
+	conn, err = client.Dial(address)
```
michalpristas: client.Dial(address)

client.go
```diff
@@ -0,0 +1,295 @@
...
+	return conn, nil
+}
+
+// DialTimeoutWithTLS acts like DialWithTLS but takes a timeout.
+func DialTimeoutWithTLS(network, address string, tlsConfig *tls.Config, timeout time.Duration) (conn *Conn, err error) {
+	if !strings.HasSuffix(network, "-tls") {
+		network += "-tls"
+	}
+	client := Client{Net: network, Dialer: &net.Dialer{Timeout: timeout}, TLSConfig: tlsConfig}
+	conn, err = client.Dial(address)
```
michalpristas: client.Dial(address)

message.go
```diff
@@ -0,0 +1,670 @@
...
+	case int:
+		v = uint32(i)
+	case int32:
+		v = uint32(i)
+	case uint:
+		v = uint32(i)
+	case uint32:
+		v = i
+	default:
+		panic(fmt.Errorf("invalid type for option %x: %T (%v)",
```
michalpristas: if this is not really needed consider returning error instead of panicing

messagetcp.go
```diff
@@ -0,0 +1,293 @@
...
+	if err := binary.Read(r, binary.BigEndian, &firstByte); err != nil {
+		return mti, normalizeErrors(err)
+	}
+	hdrOff++
+
+	lenNib := (firstByte & 0xf0) >> 4
+	tkl := firstByte & 0x0f
+
+	var opLen int
+	if lenNib < TCP_MESSAGE_LEN13_BASE {
```
michalpristas: readability: switch?

server.go
```diff
@@ -0,0 +1,720 @@
...
+		}
+	}
+
+	defer atomic.AddInt32(&srv.workersCount, -1)
+
+	inUse := false
+	timeout := time.NewTimer(idleWorkerTimeout)
+	defer timeout.Stop()
+LOOP:
+	for {
```
michalpristas: try to avoid this, maybe extract to separate func and return instead of break ANCH

server.go
```diff
@@ -0,0 +1,720 @@
...
+		// Check PacketConn interface's type is valid and value
+		// is not nil
+		if t, ok := pConn.(*net.UDPConn); ok && t != nil {
+			if e := setUDPSocketOptions(t); e != nil {
+				return e
+			}
+			srv.started = true
+			srv.lock.Unlock()
+			e := srv.serveUDP(t)
+			srv.lock.Lock() // to satisfy the defer at the top
```
michalpristas: try to think about this seems confusing

client.go
jkralik: done

README.md
jkralik: Yeah I understand. For example i removed this line because it is simple example

client.go
jkralik: Done

client.go
jkralik: Done

client.go
jkralik: Done

client.go
jkralik: Done

client.go
jkralik: Done

client.go
jkralik: Done

message.go
jkralik: Now it returns error and function upper just log fatal error for that option.

messagetcp.go
jkralik: done

server.go
jkralik: done

server.go
jkralik: done

PR #3, opened 2018-06-08
refactoring: client uses same code server

jkralik:
add Exchange to Session for challenging client
    add DefaultHandle to set default handler of mux
    use same code of Server as for Client connection
    session and client has similar interface

### DIFF
--- a/client.go
+++ b/client.go
@@ -3,48 +3,30 @@ package coap
 // A client implementation.
 
 import (
-	"bytes"
 	"crypto/tls"
 	"net"
 	"strings"
 	"time"
 )
 
-const (
-	tcpIdleTimeout time.Duration = 8 * time.Second
-)
-
-// A Conn represents a connection to a COAP server.
-type Conn struct {
-	net.Conn        // a net.Conn holding the connection
-	UDPSize  uint16 // minimum receive buffer for UDP messages
-	TCPbuf   []byte
-	client   *Client
+// A ClientConn represents a connection to a COAP server.
+type ClientConn struct {
+	srv          *Server
+	client       *Client
+	session      Session
+	shutdownSync chan error
 }
 
 // A Client defines parameters for a COAP client.
 type Client struct {
-	Net       string      // if "tcp" or "tcp-tls" (COAP over TLS) a TCP query will be initiated, otherwise an UDP one (default is "" for UDP)
-	UDPSize   uint16      // minimum receive buffer for UDP messages
-	TLSConfig *tls.Config // TLS connection configuration
-	Dialer    *net.Dialer // a net.Dialer used to set local address, timeouts and more
-	// Timeout is a cumulative timeout for dial, write and read, defaults to 0 (disabled) - overrides DialTimeout, ReadTimeout,
-	// WriteTimeout when non-zero. Can be overridden with net.Dialer.Timeout (see Client.ExchangeWithDialer and
-	// Client.Dialer) or context.Context.Deadline (see the deprecated ExchangeContext)
-	Timeout      time.Duration
-	DialTimeout  time.Duration // net.DialTimeout, defaults to 2 seconds, or net.Dialer.Timeout if expiring earlier - overridden by Timeout when that value is non-zero
-	ReadTimeout  time.Duration // net.Conn.SetReadTimeout value for connections, defaults to 2 seconds - overridden by Timeout when that value is non-zero
-	WriteTimeout time.Duration // net.Conn.SetWriteTimeout value for connections, defaults to 2 seconds - overridden by Timeout when that value is non-zero
-}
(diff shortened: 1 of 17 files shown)


client.go
```diff
@@ -91,167 +72,88 @@ func (c *Client) Dial(address string) (conn *Conn, err error) {
 		}
 	}
 
-	conn = new(Conn)
+	sync := make(chan bool)
+	clientConn = &ClientConn{srv: &Server{Net: network, TLSConfig: c.TLSConfig, ReadTimeout: c.readTimeout(), WriteTimeout: c.writeTimeout(), UDPSize: c.UDPSize,
+		NotifyStartedFunc: func() {
+			sync <- true
```
michalpristas: i would add switch with timeout logging Fatal to yell that no one is listening maybe. 
```go
switch{
 case <-sync:
 case <- time.After(3 * time.Minute): log.Fatal("something")
}
```

sample might be incorrect, it's just from top of my head

client.go
```diff
@@ -91,167 +72,88 @@ func (c *Client) Dial(address string) (conn *Conn, err error) {
...
-	// net.Dialer.Timeout has priority if smaller than the timeouts computed so
-	// far
-	if c.Dialer != nil && c.Dialer.Timeout != 0 && c.Dialer.Timeout < requestTimeout {
-		requestTimeout = c.Dialer.Timeout
-	}
-	return requestTimeout
+// Close close connection
+func (co *ClientConn) Close() {
+	co.srv.Shutdown()
+	<-co.shutdownSync
```
michalpristas: same timeout pattern, what if co fails to send? this will hang. 
with timeout switch you can give it some time to shutdown and if it fails just kill it

client.go
```diff
@@ -260,7 +162,7 @@ func DialWithTLS(network, address string, tlsConfig *tls.Config) (conn *Conn, er
 }
 
 // DialTimeoutWithTLS acts like DialWithTLS but takes a timeout.
-func DialTimeoutWithTLS(network, address string, tlsConfig *tls.Config, timeout time.Duration) (conn *Conn, err error) {
+func DialTimeoutWithTLS(network, address string, tlsConfig *tls.Config, timeout time.Duration) (conn *ClientConn, err error) {
 	if !strings.HasSuffix(network, "-tls") {
```
michalpristas: small nit: repetitive, maybe helper method

conn.go
```diff
@@ -0,0 +1,214 @@
...
+func (conn *connTCP) Close() error {
+	conn.finishWrite()
+	return conn.connection.Close()
+}
+
+func (conn *connTCP) Write(w writeReq) error {
+	if atomic.LoadInt32(&conn.closed) > 0 {
+		return ErrConnectionClosed
+	}
+	conn.writeChan <- w
```
michalpristas: again consider timeout. i'm pointing this out not as a must but consideration for you with a question. what would happen if this blocks forever

conn.go
```diff
@@ -0,0 +1,214 @@
...
+
+func newConnectionUDP(c *net.UDPConn, srv *Server) conn {
+	connection := &connUDP{connBase: connBase{writeChan: make(chan writeReq), closeChan: make(chan bool), finChan: make(chan bool), closed: 0}, connection: c}
+	go writeUDP(connection, srv)
+	return connection
+}
+
+func writeTCP(conn *connTCP, srv *Server) {
+	wr := srv.acquireWriter(conn.connection)
+LOOP:
```
michalpristas: try to avoid these, while this is perfectly fine in C and valid in Go, common practice is to extract functionality into multiple funcs and return instead of break ANCH.

client.go
jkralik: Fixed

client.go
jkralik: Fixed

client.go
jkralik: Fixed

conn.go
jkralik: added timeout as arg. Fixed

conn.go
jkralik: extracted to function: FIXED

examples/observe/client/main.go
```diff
@@ -22,21 +28,12 @@ func main() {
...
-		var rv coap.Message
-		rv, err = conn.ReadMsg(2 * time.Second)
-		if err != nil {
-			log.Fatalf("Error receiving: %v", err)
-		} else if rv != nil {
-			log.Printf("Got %s", rv.Payload())
-		}
-
-	}
+	time.Sleep(20 * time.Second)
```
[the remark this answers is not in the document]
jkralik: This is only example to show messages that arrives in 20 sec,. Of course it is problem if someone want to fix issue by sleep.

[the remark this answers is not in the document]
ondrejtomcik: Mention it explicitely in the comment so people are not scared after checking an example - as that's the first entry point to some library :)

PR #20, opened 2018-11-27
add noise to go-coap as a crypto layer.

ara4n:
done as a very quick and dirty hack, sponsored by UA900.
i think this is also my first serious go; please pull it to shreds.


ara4n:
oops, wrong repo - was aiming for my own as a first cut(!) :)

PR #22, opened 2018-12-03
Fix responding to blockwise requests using Write

erikjohnston:
When responding to blockwise requests using `ResponseWrite.Write` the server would use the message ID of the first block message rather than the last. This meant that clients would fail to handle the response.

This was due to not creating a new response writer when we created a new request.


jkralik:
Thx for find bug and fix it. :+1:

PR #27, opened 2018-12-15
Don't panic

akokhanovskyi:
Letting panics leak through the package boundary should be the last resort solution. None of the servermux conditions that lead to panic are unrecoverable.
See:
- https://github.com/golang/go/wiki/CodeReviewComments#dont-panic
- https://golang.org/doc/effective_go.html#errors
- https://golang.org/doc/effective_go.html#panic

### DIFF
--- a/servermux.go
+++ b/servermux.go
@@ -1,7 +1,10 @@
 // Package coap provides a CoAP client and server.
 package coap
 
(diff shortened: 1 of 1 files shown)


jkralik:
Thx for contributing.

PR #36, opened 2019-03-05
Provide context to Request

jkralik:
Context is accessible via `Request.Ctx` and it will be cancelled when connection was closed.
Fix deadlock when server close connection by force via call Close.
Some functionality was moved to `github.com/go-ocf/kit/net`.

This library doesn't use dep and it depends on https://github.com/go-ocf/kit - package net.


client.go
```diff
@@ -286,86 +268,109 @@ func (co *ClientConn) NewDeleteRequest(path string) (Message, error) {
...
 func (co *ClientConn) Close() error {
 	co.srv.Shutdown()
 	select {
 	case <-co.shutdownSync:
-	case <-time.After(co.client.syncTimeout()):
-		log.Fatal("Client cannot recv shutdown: Timeout")
+		/*
+			case <-time.After(co.client.syncTimeout()):
+				log.Fatal("Client cannot recv shutdown: Timeout")
+		*/
```
cizmazia: Please remove

networksession.go
```diff
@@ -395,10 +356,25 @@ func (s *sessionTCP) exchangeTimeout(req Message, writeDeadline, readDeadline ti
 		}
```
cizmazia: Please could you use `sync.Map` instead?

jkralik: For TCP yes but for UDP it is no possible because it is map in map and to delete entry i need lock whole nested objects. Then I will prefer same implementation for bot tcp and udp.

PR #37, opened 2019-03-05
support no-response option

arun1587:
OptionID has to be made uint16 in order to support the no-response
option number i.e. 258.
Added Encoding and Decoding message tests for the no-response option.
Reference: https://tools.ietf.org/html/rfc7967

### DIFF
--- a/noresponsewriter.go
+++ b/noresponsewriter.go
@@ -1,0 +1,76 @@
+package coap
+
+func isSet(n uint32, pos uint32) bool {
+	val := n & (1 << pos)
+	return (val > 0)
+}
+
+func powerOfTwo(exponent uint32) uint32 {
+	if exponent != 0 {
+		return (2 * powerOfTwo(exponent-1))
+	}
+	return 1
+}
+
+func (w *noResponseWriter) decodeNoResponseOption(v uint32) []COAPCode {
+	var codes []COAPCode
+	if v == 0 {
+		// No suppresed code
+		return codes
+	}
+
+	var i uint32
+	// Max possible value:16; ref:table_2_rfc7967
+	for i = 0; i <= 4; i++ {
+		if isSet(v, i) {
+			index := powerOfTwo(i)
+			codes = append(codes, w.noResponseValueMap[index]...)
+		}
+	}
+	return codes
+}
+
+type noResponseWriter struct {
+	*responseWriter
+	noResponseValueMap map[uint32][]COAPCode
+}
+
+func newNoResponseWriter(w ResponseWriter) *noResponseWriter {
+	return &noResponseWriter{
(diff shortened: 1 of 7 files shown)


jkralik:
Thank you for contributing! Please could you:
- update file README.MD about this RFC
- create NoResponseWriter that satisfy interface -> it will writes only message with allowed code, otherwise it returns error.

noresponsewriter.go
```diff
@@ -0,0 +1,76 @@
...
+		if isSet(v, i) {
+			index := powerOfTwo(i)
+			codes = append(codes, w.noResponseValueMap[index]...)
+		}
+	}
+	return codes
+}
+
+type noResponseWriter struct {
+	*responseWriter
```
jkralik: please replace `*responseWriter` with `responseWriter ResponseWriter`

noresponsewriter.go
```diff
@@ -0,0 +1,76 @@
...
+	noResponseValueMap map[uint32][]COAPCode
+}
+
+func newNoResponseWriter(w ResponseWriter) *noResponseWriter {
```
jkralik: please use factory at line https://github.com/go-ocf/go-coap/blob/3638c5d5aefdb54c24ba9ec02dddf113bed6a49c/responsewriter.go#L64

noresponsewriter.go
```diff
@@ -0,0 +1,76 @@
...
+}
+
+func (w *noResponseWriter) Write(p []byte) (n int, err error) {
+	l, resp := prepareReponse(w, w.responseWriter.req.Msg.Code(), w.responseWriter.code, w.responseWriter.contentFormat, p)
+	err = w.WriteMsg(resp)
+	return l, err
+}
+
+func (w *noResponseWriter) SetCode(code COAPCode) {
+	w.code = &code
```
jkralik: pls forward call to responseWriter `w.responseWriter.SetCode(code)`

arun1587:
> Please add test for noResponseWriter, add update MR against new go-coap.
> Thx
I rebased on recent changes. added tests for decoding but can you give any reference for improving the coverage, especially the methods of `ResponseWriter`.

jkralik:
Create server via `RunLocalServerUDPWithHandler("udp", ":", false,  BlockWiseSzx16, func (w ResponseWriter, r *Request) { testNoResponseHandler(t, w, r }) `  - It returns server, addr where client need to connect and fin channel (when server you call shutdown then read fin chan).  And connect client, create req via New{Get/Post/...}Request, set option NoResponse and test with WriteMsg/ExchangeMsg.

arun1587:
You could squash the commits before merge.,
Thanks.

jkralik:
Good works !

PR #68, opened 2019-10-27
Go Modules support

### LINKED ISSUE #66: Go Modules support
Hi there!

With the introduction of Go 1.11 Modules were introduced. Do you plan to support those?

I can happily submit a PR. It's usually as easy as running
```sh
go mod init github.com/axiomhq/quantiles
go mod tidy
```

maciej:
Coverts package management definitions from dep to Go Module
along with supporting docker build.

It also bumps Go in Dockerfile by one minor version. This change
is non-essential.

Resolves #66


Dockerfile
```diff
@@ -1,11 +1,5 @@
...
+FROM golang:1.12-alpine3.10 AS build
 
-RUN apk add --no-cache curl git build-base && \
-	curl -SL -o /usr/bin/dep https://github.com/golang/dep/releases/download/v0.5.0/dep-linux-amd64 && \
-	chmod +x /usr/bin/dep
-
-ENV MAINDIR $GOPATH/src/github.com/go-ocf/go-coap
-WORKDIR $MAINDIR
-COPY Gopkg.toml Gopkg.lock ./
-RUN dep ensure -v --vendor-only
```
cizmazia: I would like to have this optimization preserved for Go Modules to allow Docker to cache downloaded dependencies when the `go.mod` file is unchanged.

maciej: Good point. I'll fix that.

maciej:
@cizmazia I've updated the PR. Would you mind having another look at it?

Dockerfile
ivajloip: Hi,

I am not sure the current version will allow Docker to cache the download as copying `.` will generally change quite often invalidating the cache of any steps afterwards. I believe you should just copy `go.mod` and `go.sum` first, do the `go mod download` and then copy the rest. This way if `go.mod` and `go.sum` didn't change, the downloading will still be taken from the cache.

cizmazia:
```dockerfile
COPY go.mod go.sum ./
RUN go mod download
COPY . .
````

maciej: @ivajloip @cizmazia good point!

maciej: Updated.

jkralik:
Thx for contributing!!!

PR #74, opened 2019-11-11
surpport unkown options

duanxuelin:
For options those are not defined by IETF, old implementation of the program is ignore them.  
Updated `parseOptionValue`  so we can get original value.  
related issue [https://github.com/go-ocf/go-coap/issues/71](71)


jkralik:
Hi. I merge it to different branch where I fixed test TestDecodeMessageWithUnknownOption. Thx for contributing!

PR #78, opened 2019-11-13
ListenAndServe should only return on shutdown or fatal error

hjames9:
@jkralik 
Regarding issue https://github.com/go-ocf/go-coap/issues/67

The review has a lot of noise due to moving net package into coap.  The reason I did that was because I made use of error objects in error.go that's in the coap package and seems like it should be accessible as "coap.ErrServerClosed".  Similar to the golang http library "http.ErrServerClosed".  Let me know if you want to restructure a different way instead.

The important changes are in dtlslistener.go, tlslistener.go, tcplistener.go where it returns ErrServerClosed on the context.Done() channel on shutdown.  Then inside of the accept loops for DTLS, TCP, etc, inside of server.go, instead of returning on every error, we continue and only return when an ErrServerClosed error is encountered.

### DIFF
--- a/server.go
+++ b/server.go
@@ -532,8 +532,13 @@ func (srv *Server) serveDTLSListener(l Listener) error {
 	for {
 		rw, err := l.AcceptWithContext(ctx)
 		if err != nil {
-			wg.Wait()
-			return fmt.Errorf("cannot serve dtls: %v", err)
+			switch err {
+			case ErrServerClosed:
+				wg.Wait()
+				return fmt.Errorf("cannot serve dtls: %v", err)
+			default:
+				continue
+			}
 		}
 		if rw != nil {
 			wg.Add(1)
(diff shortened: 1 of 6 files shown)


server.go
```diff
@@ -602,13 +606,18 @@ func (srv *Server) serveTCPListener(l Listener) error {
 		rw, err := l.AcceptWithContext(ctx)
 		if err != nil {
 			wg.Wait()
```
jkralik: `wg` must be moved under `case ErrServerClosed` -> it used to wait for go-routines that served connections.

server.go
```diff
@@ -533,19 +532,24 @@ func (srv *Server) serveDTLSListener(l Listener) error {
...
```
jkralik: `wg` must be moved under `case ErrServerClosed` -> it used to wait for go-routines that served connections.

hjames9:
@jkralik This is a lot simpler and cleaner now

jkralik:
Good job!

PR #93, opened 2020-02-05
implement keepalive feature

jkralik:


backoff.go
```diff
@@ -0,0 +1,11 @@
...
+
+import (
+	"github.com/cenkalti/backoff/v4"
+)
+
+// Stop indicates that no more retries should be made for use in NextBackOff().
+const Stop = backoff.Stop
+
+// BackOff is a backoff policy for retrying an operation.
+type BackOff = backoff.BackOff
```
cizmazia: I would prefer for the policy to be represented using the native types. The last time, we have agreed to wrap the policy in this interface:
```go
type RetryFunc = func() (when time.Time, err error)
type RetryFuncFactory = func() RetryFunc

func handleRetry(ctx context.Context, retryFunc RetryFunc) error {
	when, err := retryFunc()
	if err != nil {
		return fmt.Errorf("cannot retry: %v", err)
	}
	select {
	case <-time.After(time.Until(when)):
	case <-ctx.Done():
		return fmt.Errorf("retry canceled")
	}
	return nil
}
```

keepalive_test.go
```diff
@@ -0,0 +1,392 @@
...
+	s, addr, _, err := RunLocalTLSServer(":0", &config)
+	require.NoError(t, err)
+	defer s.Shutdown()
+
+	var wg sync.WaitGroup
+	wg.Add(1)
+	c := &Client{
+		Net:       "tcp-tls",
+		TLSConfig: &tls.Config{InsecureSkipVerify: true},
+		KeepAlive: KeepAlive{
```
cizmazia: Please could you provide a factory function:

```go
// MakeKeepAlive creates a policy that detects dropped connections within the connTimeout limit
func MakeKeepAlive(connTimeout time.Duration) (KeepAlive, error)
```

server.go
```diff
@@ -319,21 +321,45 @@ func (srv *Server) ListenAndServe() error {
...
 		srv.NotifyStartedFunc()
 	}
-	return srv.serveTCPConnection(newShutdownWithContext(srv.doneChan), conn)
+	return srv.serveTCPConnection(newShutdownWithContext(doneChan), conn)
 }
 
 func (srv *Server) initServeDTLS(conn *coapNet.Conn) error {
+	srv.doneLock.Lock()
+	if srv.doneChan != nil {
+		return fmt.Errorf("server already serve connections")
```
cizmazia: Please could you always use `defer` for unlocking? This path does not unlock. Consider using [Event](https://godoc.org/google.golang.org/grpc/internal/grpcsync).

server.go
```diff
@@ -319,21 +321,45 @@ func (srv *Server) ListenAndServe() error {
 }
 
 func (srv *Server) initServeUDP(connUDP *coapNet.ConnUDP) error {
-	return srv.serveUDP(newShutdownWithContext(srv.doneChan), connUDP)
+	srv.doneLock.Lock()
+	if srv.doneChan != nil {
+		return fmt.Errorf("server already serve connections")
```
cizmazia: Please could you always use defer for unlocking? This path does not unlock. Consider using [Event](https://godoc.org/google.golang.org/grpc/internal/grpcsync).

server.go
```diff
@@ -597,12 +640,20 @@ func (srv *Server) serveTCPConnection(ctx *shutdownContext, conn *coapNet.Conn)
 
 // serveListener starts a TCP listener for the server.
 func (srv *Server) serveTCPListener(l Listener) error {
+	srv.doneLock.Lock()
+	if srv.doneChan != nil {
+		return fmt.Errorf("server already serve connections")
```
cizmazia: Please could you always use `defer` for unlocking? This path does not unlock. Consider using [Event](https://godoc.org/google.golang.org/grpc/internal/grpcsync).

keepalive.go
```diff
@@ -19,7 +22,42 @@ type KeepAlive struct {
...
+		NewRetryPolicy: func() RetryFunc {
+			now := time.Now()
+			// try 2 times to send ping after fails (wait, pong, wait, pong)
+			return func() (time.Time, error) {
+				c := time.Now()
+				if c.Before(now.Add(duration * 4)) {
+					return c.Add(duration), nil
+				}
+				return time.Time{}, ErrKeepAliveDeadlineExceeded
+			}
```
cizmazia: To prevent drifting:

```go
start := time.Now()
attempt := 0
return func() (time.Time, error) {
	attempt++
	// Try to send ping and wait for pong 2 more times
	if time.Since(start) <= 2 * 2 * duration {
		return start.Add(attempt * 2 * duration), nil
	}
	return time.Time{}, ErrKeepAliveDeadlineExceeded
}
```

keepalive_test.go
```diff
@@ -29,12 +28,15 @@ func TestKeepAliveTCP_Client(t *testing.T) {
 			Enable:      true,
 			WaitForPong: time.Microsecond,
 			Interval:    time.Millisecond * 100,
-			NewRetryPolicy: func() BackOff {
-				bo := backoff.NewExponentialBackOff()
-				bo.InitialInterval = 200 * time.Millisecond
-				bo.MaxInterval = 1 * time.Second
-				bo.MaxElapsedTime = 2 * time.Second
-				return bo
+			NewRetryPolicy: func() RetryFunc {
```
cizmazia: Could you use `MustMakeKeepAlive` to cover it and avoid repetition?

server.go
```diff
@@ -321,41 +321,42 @@ func (srv *Server) ListenAndServe() error {
...
 		srv.NotifyStartedFunc()
 	}
 	return srv.serveTCPConnection(newShutdownWithContext(doneChan), conn)
 }
 
-func (srv *Server) initServeDTLS(conn *coapNet.Conn) error {
+func (srv *Server) initDone() (<-chan struct{}, error) {
 	srv.doneLock.Lock()
+	defer srv.doneLock.Unlock()
 	if srv.doneChan != nil {
```
cizmazia: Why to use `chan` and lock instead of the higher level [Event](https://godoc.org/google.golang.org/grpc/internal/grpcsync)?

keepalive_test.go
jkralik: No because MustMakeKeepAlive  returns reasonable value but I want to simulate when keepalive fails.

cizmazia: I am sure you can use `MustMakeKeepAlive` for the success case, right?

server.go
jkralik: lock -> because shutdown can be called by any goroutines and assignation Event/channel  must be done under mutex.

cizmazia: `Event` already uses a `mutex` internally.

keepalive_test.go
jkralik: The lastest test of keepalive used `MustMakeKeepAlive `.

PR #97, opened 2020-02-23
add support to stringify coap options

atreya2011:
Thank you for merging my previous pull request. In this pull request, I have added support to stringify CoAP options which was useful to me when I wanted to parse a CoAP message and print out the CoAP option name along with its value present within the message. Looking forward to your review and subsequent merge to master! 🙏

### DIFF
--- a/message.go
+++ b/message.go
@@ -149,6 +149,53 @@ var coapOptionDefs = map[OptionID]optionDef{
 	NoResponse:    optionDef{valueFormat: valueUint, minLen: 0, maxLen: 1},
 }
 
+func (o OptionID) String() string {
+	switch o {
+	case IfMatch:
+		return "IfMatch"
+	case URIHost:
+		return "URIHost"
+	case ETag:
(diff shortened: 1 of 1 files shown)


message.go
```diff
@@ -149,6 +149,33 @@ var coapOptionDefs = map[OptionID]optionDef{
...
+	Block2:        "Block2",
+	Block1:        "Block1",
+	Size2:         "Size2",
+	ProxyURI:      "ProxyURI",
+	ProxyScheme:   "ProxyScheme",
+	Size1:         "Size1",
+	NoResponse:    "NoResponse",
+}
+
+func (o OptionID) String() string {
```
jkralik: Pls could you change function to similar like https://github.com/go-ocf/go-coap/blob/master/codes/code_string.go ? Because we want to print unknown option.

atreya2011: Thank you for your review. I have made the change as you requested. 👍

PR #106, opened 2020-03-25
feat(cert): add peer certificates to the incoming request

### LINKED ISSUE #103: Support for DTLS certificate export
I couldn't find this to be supported ATM, but how difficult would it be to support this or is this already supported?

I'm also still figuring out in pion/dtls how i can export certificates, so once that is figured out, this ticket should become more specific. Nonetheless, I figured creating a ticket here to see if anyone has done anything similar, and get the ball rolling once I've figured out how to export the certificate from pion/dtls.
After a fist look (and I'm not sure whether this is indeed right until the dtls is figured out) it might mean that the `dtls.Conn` needs to be exposed somehow from `coap.ListenAndServeDTLS`.

jdbruijn:
This will, for example, allow the server to get the client certificate (chain) in case of DTLS.
The certificates can be accessed as `req.PeerCertificates` where `req` is of type `req *coap.Request`.

Fixes #103

@jkralik Could you please look at these changes? I'd love some feedback and am happy to change this in any way. The changes I've done now worked for me, but I'm not sure whether the way I've done these changes is preferred as I'm new to the codebase (and Go).

Some elements in specific I'm not sure about:
- In [net/conn.go](https://github.com/jdbruijn/go-coap/blob/bedcb6752729174f0306f22296298c4848a141fa/net/conn.go#L38) I'm exposing the `Conn`, which could be altered by the receiver if I'm not mistaken. Although the same goes for the `LocalAddr` and `RenoteAddr` functions, I'm not sure whether this is okay for the `Conn` too.
- Ignoring possible errors in the certificate parsing [here](https://github.com/jdbruijn/go-coap/blob/bedcb6752729174f0306f22296298c4848a141fa/server.go#L564). I've added a comment about it but if you look at it differntly, please let me know and I'll change it.

Attached is the patch file which I used to verify the workings of this since PSK doesn't have certificates of course.

[peer-certificates-test.patch.txt](https://github.com/go-ocf/go-coap/files/4380263/peer-certificates-test.patch.txt)

BTW: This time I've also ran the tests locally, so should be passing ;)

### DIFF
--- a/request.go
+++ b/request.go
@@ -2,13 +2,11 @@ package coap
 
 import (
 	"context"
-	"crypto/x509"
 )
 
 type Request struct {
-	Msg              Message
-	PeerCertificates []*x509.Certificate
-	Client           *ClientConn
-	Ctx              context.Context
-	Sequence         uint64 // discontinuously growing number for every request from connection starts from 0
+	Msg      Message
+	Client   *ClientConn
+	Ctx      context.Context
+	Sequence uint64 // discontinuously growing number for every request from connection starts from 0
 }
--- a/request.go
+++ b/request.go
@@ -1,10 +1,14 @@
 package coap
 
-import "context"
+import (
+	"context"
+	"crypto/x509"
+)
 
 type Request struct {
-	Msg      Message
-	Client   *ClientConn
-	Ctx      context.Context
-	Sequence uint64 // discontinuously growing number for every request from connection starts from 0
+	Msg              Message
+	PeerCertificates []*x509.Certificate
+	Client           *ClientConn
+	Ctx              context.Context
+	Sequence         uint64 // discontinuously growing number for every request from connection starts from 0
 }
--- a/server.go
+++ b/server.go
@@ -2,8 +2,10 @@
 package coap
 
 import (
+	"bytes"
 	"context"
 	"crypto/tls"
+	"crypto/x509"
 	"fmt"
 	"net"
 	"reflect"
@@ -553,7 +555,15 @@ func (srv *Server) serveDTLSConnection(ctx *shutdownContext, conn *coapNet.Conn)
 		// We will block poller wait loop when
 		// all pool workers are busy.
 		c := ClientConn{commander: &ClientCommander{session}}
-		srv.spawnWorker(&Request{Client: &c, Msg: msg, Ctx: sessCtx, Sequence: c.Sequence()})
+
+		// Try to set the peer certificates, but don't error if it fails as there
+		// might not be any peer certificates.
+		dtlsConn := conn.Connection().(*dtls.Conn)
+		cert := dtlsConn.RemoteCertificate()
+		flatCerts := bytes.Join(cert, nil)
+		certs, _ := x509.ParseCertificates(flatCerts)
+
+		srv.spawnWorker(&Request{Client: &c, PeerCertificates: certs, Msg: msg, Ctx: sessCtx, Sequence: c.Sequence()})
 	}
 }
 
--- a/server.go
+++ b/server.go
@@ -2,10 +2,8 @@
 package coap
 
 import (
-	"bytes"
 	"context"
 	"crypto/tls"
-	"crypto/x509"
 	"fmt"
 	"net"
 	"reflect"
@@ -555,15 +553,7 @@ func (srv *Server) serveDTLSConnection(ctx *shutdownContext, conn *coapNet.Conn)
 		// We will block poller wait loop when
 		// all pool workers are busy.
 		c := ClientConn{commander: &ClientCommander{session}}
-
-		// Try to set the peer certificates, but don't error if it fails as there
-		// might not be any peer certificates.
-		dtlsConn := conn.Connection().(*dtls.Conn)
-		cert := dtlsConn.RemoteCertificate()
-		flatCerts := bytes.Join(cert, nil)
-		certs, _ := x509.ParseCertificates(flatCerts)
-
-		srv.spawnWorker(&Request{Client: &c, PeerCertificates: certs, Msg: msg, Ctx: sessCtx, Sequence: c.Sequence()})
+		srv.spawnWorker(&Request{Client: &c, Msg: msg, Ctx: sessCtx, Sequence: c.Sequence()})
 	}
 }
 
--- a/sessiondtls.go
+++ b/sessiondtls.go
@@ -33,18 +33,21 @@ func newSessionDTLS(connection *coapNet.Conn, srv *Server) (networkSession, erro
 		return nil, ErrInvalidBlockWiseSzx
 	}
 
-	dtlsConn := connection.Connection().(*dtls.Conn)
-	cert := dtlsConn.RemoteCertificate()
-	flatCerts := bytes.Join(cert, nil)
-	peerCertificates, err := x509.ParseCertificates(flatCerts)
-	if err != nil {
-		return nil, err
+	s := sessionDTLS{
+		sessionBase: newBaseSession(BlockWiseTransfer, BlockWiseTransferSzx, srv),
+		connection:  connection,
 	}
 
-	s := sessionDTLS{
-		sessionBase:      newBaseSession(BlockWiseTransfer, BlockWiseTransferSzx, srv),
-		connection:       connection,
-		peerCertificates: peerCertificates,
+	dtlsConn := connection.Connection().(*dtls.Conn)
+	cert := dtlsConn.RemoteCertificate()
+	if len(cert) > 0 {
+		flatCerts := bytes.Join(cert, nil)
+		peerCertificates, err := x509.ParseCertificates(flatCerts)
+		if err != nil {
+			return nil, err
+		}
+
+		s.peerCertificates = peerCertificates
 	}
 
 	return &s, nil
--- a/sessiondtls.go
+++ b/sessiondtls.go
@@ -3,16 +3,19 @@ package coap
 import (
 	"bytes"
 	"context"
+	"crypto/x509"
 	"fmt"
 	"net"
 
 	"github.com/go-ocf/go-coap/codes"
 	coapNet "github.com/go-ocf/go-coap/net"
+	dtls "github.com/pion/dtls/v2"
 )
 
 type sessionDTLS struct {
 	*sessionBase
-	connection *coapNet.Conn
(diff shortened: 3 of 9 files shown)


request.go
```diff
@@ -1,10 +1,14 @@
...
+	"crypto/x509"
+)
 
 type Request struct {
-	Msg      Message
-	Client   *ClientConn
-	Ctx      context.Context
-	Sequence uint64 // discontinuously growing number for every request from connection starts from 0
+	Msg              Message
+	PeerCertificates []*x509.Certificate
```
jkralik: I think PeerCertificates doesn't bellows to Request.  It will be better as receiver function in ClientConn (just forward calls via commenader to the session, and then to the connection.

server.go
```diff
@@ -553,7 +555,15 @@ func (srv *Server) serveDTLSConnection(ctx *shutdownContext, conn *coapNet.Conn)
...
 		// all pool workers are busy.
 		c := ClientConn{commander: &ClientCommander{session}}
-		srv.spawnWorker(&Request{Client: &c, Msg: msg, Ctx: sessCtx, Sequence: c.Sequence()})
+
+		// Try to set the peer certificates, but don't error if it fails as there
+		// might not be any peer certificates.
+		dtlsConn := conn.Connection().(*dtls.Conn)
+		cert := dtlsConn.RemoteCertificate()
+		flatCerts := bytes.Join(cert, nil)
+		certs, _ := x509.ParseCertificates(flatCerts)
```
jkralik: This parts can be implemented in function newSessionDTLS

jkralik:
Thx for contributing.
> In net/conn.go I'm exposing the Conn, which could be altered by the receiver if I'm not mistaken. Although the same goes for the LocalAddr and RenoteAddr functions, I'm not sure whether this is okay for the Conn too.
It's OK.  Owner of connection is net/Conn.Conn and for one instance of conn you get same result - same RemoteAddr, same LocalAddr, Connection.

> Ignoring possible errors in the certificate parsing here. I've added a comment about it but if you look at it differntly, please let me know and I'll change it.
From my point view I would like to have checks on error.

jdbruijn:
Thanks a lot for your feedback!

I'm currently working on the changes.

server.go
jdbruijn: I've got the forwarding part via the commander implemeneted now! This part should be moved to the new `(s *sessionDTLS) RemoteCertificates()` function, right? If I move this to `newSessionDTLS`, then we'd need to store the certificates in the `sessionDTLS`.

That might be even nicer, what do you think?
1. The `RemoteCertificates` function gets and parses the certificates. It would becomes something like this, but with added error checking on the certificate.

   ```go
	func (s *sessionDTLS) RemoteCertificates() ([]*x509.Certificate, error) {
		dtlsConn := s.connection.Connection().(*dtls.Conn)
		cert := dtlsConn.RemoteCertificate()
		flatCerts := bytes.Join(cert, nil)
		certs, _ := x509.ParseCertificates(flatCerts)
		return certs, nil
	}

   ```
2. `newSessionDTLS` gets and parses the certificates and stores them in `sessionDTLS`. This way the `RemoteCertificates` just becomes a 'getter` function. and wouldn't need to return an error. I'd be adding the code to get the certificates [here](https://github.com/go-ocf/go-coap/blob/master/sessiondtls.go#L32), and return an error from `newSessionDTLS` in case of an parse error.

jdbruijn: I've just pushed af98d1a, where I went for option 2 since I think that would be the 'cleanest' solution. If you, however, rather see it another way, please let me know and I'll update it.

jdbruijn:
I've just pushed the changes and think I've covered everything form your review. Thanks again for taking the time to review this!

If I missed something, please let me know!

sessiondtls.go
```diff
@@ -30,9 +33,18 @@ func newSessionDTLS(connection *coapNet.Conn, srv *Server) (networkSession, erro
 		return nil, ErrInvalidBlockWiseSzx
 	}
 
+	dtlsConn := connection.Connection().(*dtls.Conn)
+	cert := dtlsConn.RemoteCertificate()
+	flatCerts := bytes.Join(cert, nil)
+	peerCertificates, err := x509.ParseCertificates(flatCerts)
+	if err != nil {
+		return nil, err
+	}
```
jkralik: Pls add the code under `if len(cert) > 0 { ... }` because we want to support DTLS with PSK

jkralik:
Pls could you add tests to https://github.com/go-ocf/go-coap/blob/master/dtls_test.go ?

sessiondtls.go
jdbruijn: Changed in ff8d51a. I had to move it below creation of  `sessionDTLS` in order to be able to set the `s.peerCertificates` based on whether we have parsed certificates or not.

jdbruijn:
> Pls could you add tests to https://github.com/go-ocf/go-coap/blob/master/dtls_test.go ?

Sure! I've looked at the test, but don't understand how I can add a test that checks retrieving the certificate. I think it would need t look something like `TestRSACerts`, but I don't see how I can e.g. verify that I can receive `r.Client.PeerCertificates()`.

Do I just create a copy of `TestRSACerts` and name it `TestRSAPeerCerts` and also copy `EchoServer` to `EchoServerWithPeerCerts`? But then how would I create an error if the `PeerCertificates` is empty when it shoudn't be?

jkralik:
Just creates similar test as is TestRSACerts in dtls_test.go for ECDSA and check PeerCertificates in Client and Server.Handler.

jdbruijn:
> Just creates similar test as is TestRSACerts in dtls_test.go for ECDSA and check PeerCertificates in Client and Server.Handler.

I've tried it but unfortunately don't think I quite get it yet.
I've created the following patch ([dtls_test.patch.txt](https://github.com/go-ocf/go-coap/files/4407203/dtls_test.patch.txt)), but when I run these tests, it is taking forever and looks like to be stuck on `s.ListenAndServe()`. And that is still with the `EchoServer` and not yet with the new `EchoServerPeerCertificates` I've created as `Server.Handler` (of which I can't figure out how to error when `PeerCertificates` is not set, but that is probably for later as the test doesn't work at all ATM).
The `s.ListenAndServe` is supposed to error given the `require.Error(t, err)` statement below it, but I think it doesn't error as the test keeps running forever without output. Did I do something wrong in the test/config here?

```diff
diff --git a/dtls_test.go b/dtls_test.go
index ba6f4fe..08c4351 100644
--- a/dtls_test.go
+++ b/dtls_test.go
@@ -2,6 +2,7 @@ package coap
 
 import (
 	"crypto/ecdsa"
+	"crypto/elliptic"
 	"crypto/rand"
 	"crypto/rsa"
 	"crypto/tls"
@@ -99,6 +100,28 @@ func generateRSACertsPEM(t *testing.T) ([]byte, []byte) {
 	return pem.EncodeToMemory(&pem.Block{Type: "CERTIFICATE", Bytes: derBytes}), pem.EncodeToMemory(pemBlockForKey(priv))
 }
 
+func generateECDSACertsPEM(t *testing.T) ([]byte, []byte) {
+	priv, err := ecdsa.GenerateKey(elliptic.P256(), rand.Reader)
+	if err != nil {
+		require.NoError(t, err)
+	}
+	template := x509.Certificate{
+		SerialNumber: big.NewInt(1),
+		Subject: pkix.Name{
+			Organization: []string{"Acme CEncodeToMemoryo"},
+		},
+		NotBefore: time.Now(),
+		NotAfter:  time.Now().Add(time.Hour * 24 * 180),
+
+		KeyUsage:              x509.KeyUsageKeyEncipherment | x509.KeyUsageDigitalSignature,
+		ExtKeyUsage:           []x509.ExtKeyUsage{x509.ExtKeyUsageServerAuth},
+		BasicConstraintsValid: true,
+	}
+	derBytes, err := x509.CreateCertificate(rand.Reader, &template, &template, publicKey(priv), priv)
+	require.NoError(t, err)
+	return pem.EncodeToMemory(&pem.Block{Type: "CERTIFICATE", Bytes: derBytes}), pem.EncodeToMemory(pemBlockForKey(priv))
+}
+
 func TestRSACerts(t *testing.T) {
 	cert, key := generateRSACertsPEM(t)
 	c, err := tls.X509KeyPair(cert, key)
@@ -120,3 +143,25 @@ func TestRSACerts(t *testing.T) {
 	err = s.ListenAndServe()
 	require.Error(t, err)
 }
+
+func TestECDSACertsWithPeerCertificates(t *testing.T) {
+	cert, key := generateECDSACertsPEM(t)
+	c, err := tls.X509KeyPair(cert, key)
+	require.NoError(t, err)
+	t.Log(cert, key)
+
+	config := &dtls.Config{
+		Certificates:         []tls.Certificate{c},
+		ExtendedMasterSecret: dtls.RequireExtendedMasterSecret,
+		ClientAuth:           dtls.RequireAnyClientCert,
+	}
+
+	s := &Server{
+		Net:        "udp-dtls",
+		Addr:       ":5688",
+		DTLSConfig: config,
+		Handler:    HandlerFunc(EchoServer),
+	}
+	err = s.ListenAndServe()
+	require.Error(t, err)
+}
diff --git a/server_test.go b/server_test.go
index 5893f58..e3fb567 100644
--- a/server_test.go
+++ b/server_test.go
@@ -55,6 +55,20 @@ func EchoServer(w ResponseWriter, r *Request) {
 	}
 }
 
+func EchoServerPeerCertificates(w ResponseWriter, r *Request) {
+	// if len(r.Client.PeerCertificates()) != 1 {
+	// 	fmt.Printf("Peer certificates are invalid")
+	// 	return
+	// }
+
+	if r.Msg.IsConfirmable() {
+		err := w.WriteMsg(CreateRespMessageByReq(r.Client.networkSession().IsTCP(), codes.Valid, r.Msg))
+		if err != nil {
+			fmt.Printf("Cannot write echo %v", err)
+		}
+	}
+}
+
 func EchoServerBadID(w ResponseWriter, r *Request) {
 	if r.Msg.IsConfirmable() {
 		w.WriteMsg(CreateRespMessageByReq(r.Client.networkSession().IsTCP(), codes.BadRequest, r.Msg))
```

jkralik:
Hi. I created new branch where I added test for PeerCertificate: https://github.com/go-ocf/go-coap/pull/115. Pls look at it and if it is OK i will be merge it to master.

Issue #67, opened 2019-10-27
Connection from client not reestablished when server restarts using DTLS


hjames9:
Using server from this example:
https://github.com/go-ocf/go-coap/blob/master/examples/dtls/server/main.go
i.e. "go run main.go"

And slightly modified client to continuously call coap.Get() every 3 seconds to send/receive data
https://github.com/go-ocf/go-coap/blob/master/examples/dtls/client/main.go
```
    co, err := coap.DialDTLS("udp", "localhost:5688", &dtls.Config{
        PSK: func(hint []byte) ([]byte, error) {
            fmt.Printf("Server's hint: %s \n", hint)
            return []byte{0xAB, 0xC1, 0x23}, nil 
        },
        PSKIdentityHint: []byte("Pion DTLS Server"),
        CipherSuites:    []dtls.CipherSuiteID{dtls.TLS_PSK_WITH_AES_128_CCM_8},
    })  
    if err != nil {
        log.Fatalf("Error dialing: %v", err)
    }   
    for {
        path := "/a"
        if len(os.Args) > 1 { 
            path = os.Args[1]
        }
        resp, err := co.Get(path)
        if err != nil {
            log.Fatalf("Error sending request: %v", err)
        }

        log.Printf("Response payload: %v", resp.Payload())
        time.Sleep(3 * time.Second)
    }
```

If I restart the server process, the client never recovers and detects that the Get() fails.  Get() blocks indefinitely and you're never able to determine that the connection fails and attempt to reestablish it via Dial().  I'm unsure if I'm missing something, but what's the best practice to follow in this scenario?

hjames9:
Another issue (related, but problem with the server instead of the client).  When a server goes down and comes back, with a client being unaware and continues to send messages, the coap.ListenAndServeDTLS() on the server returns with failure "cannot serve tcp: cannot accept connections: dtls: The connection timed out during the handshake"

If you attempt to call coap.ListenAndServeDTLS() on the same port, after it returns, it fails binding the port:  "cannot listen and serve: cannot create new dtls listener: listen udp :5688: bind: address already in use"

Looks like two issues.
1. Doesn't seem like ListenAndServerDTLS() should return failure if one client happens to be in a different state and is sending messages (perhaps the handshaking process shouldn't be triggered at all here?)
2. If/When ListenAndServerDTLS() does return, it doesn't seem as if it's properly cleaning up the port it's listening on.

jkralik:
Hi

[...]

PR #122, opened 2020-05-06
V2

jkralik:
- protocols are separated to packages
- observe work's natively (https://github.com/go-ocf/go-coap/issues/12, https://github.com/go-ocf/go-coap/issues/99)
- support's separate message
- don't propagate reset message to client(https://github.com/go-ocf/go-coap/issues/110)
- body as ReadSeeker for easy transfer files
- ability to set source port (https://github.com/go-ocf/go-coap/issues/90)
- (un)marshal udp/tcp coap message without allocation
- added support for middleware to mux #73

PR #123, opened 2020-05-11
Add callback for errors occurring from listeners and fix DTLS accept loop from exiting

hjames9:
@jkralik As per issue #67, implemented callbacks into listeners to return back error messages on any failures.  Also fixed a bug where the DTLS accept loop would exit if a single session has a handshake failure

### DIFF
--- a/dtls_test.go
+++ b/dtls_test.go
@@ -128,11 +128,17 @@ func TestRSACerts(t *testing.T) {
 		ClientAuth:           dtls.RequireAnyClientCert,
 	}
 
+	listenerErrorHandler := func(err error) bool {
+		fmt.Fprintf(os.Stderr, "Listener error occurred: %v", err)
+		return true
+	}
+
 	s := &Server{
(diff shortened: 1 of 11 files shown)


hjames9:
@jkralik Let me know if this interface for the listener error callback suits you, happy to change it if you'd prefer something different.  Thanks!

jkralik:
Hi @hjames9 . 

Thx for contributing! I will accept your whole PR to V0 but for V2(#122 ) I will take only part of  `fix DTLS accept loop from exiting`, because I cleaned up API and Listener is a argument of Serve function: https://github.com/go-ocf/go-coap/blob/84c337551b2ae0f96ad86b7d0ca1c5ca776bf628/dtls/server.go#L152  and you can wrap `DTLSListener.AcceptWithContext`  as it is needed.

V2: is still under progress but in near future it will be merged to master.

hjames9:
@jkralik Thanks a lot, makes perfect sense

PR #132, opened 2020-06-16
fix of close sessions

jkralik:
- addOnClose addOnRun with mutex
- handle ctx.Done in udp client/server

### DIFF
--- a/udp/session.go
+++ b/udp/session.go
@@ -18,12 +18,12 @@ type Session struct {
 	raddr          *net.UDPAddr
 	maxMessageSize int
 
+	mutex   sync.Mutex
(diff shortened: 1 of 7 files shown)


udp/session.go
```diff
@@ -47,16 +47,47 @@ func (s *Session) Done() <-chan struct{} {
 }
 
 func (s *Session) AddOnClose(f EventFunc) {
+	s.mutex.Lock()
+	defer s.mutex.Unlock()
 	s.onClose = append(s.onClose, f)
 }
 
 func (s *Session) AddOnRun(f EventFunc) {
```
cizmazia, line 55: Remove if redundant

PR #149, opened 2020-08-05
fix: use loop index to calculate used length of options

jdbruijn:
The wrong index variable in the loop was used, potentially causing not enough bytes allocated for the next cycle of the mux options conversion.

### DIFF
--- a/message/options.go
+++ b/message/options.go
@@ -536,7 +536,7 @@ func (options Options) ResetOptionsTo(buf []byte, in Options) (Options, int, err
 	for idx, o := range in {
 		if len(buf) < len(o.Value) {
 			for i := idx; i < len(in); i++ {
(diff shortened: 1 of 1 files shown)


jkralik:
Thx for contributing. 👍

jkralik:
I will grant you write access to go-ocf/go-coap.

jdbruijn:
Thanks for the quick processing!

> I will grant you write access to go-ocf/go-coap.

Thanks. Just to check what the intention with this is, is this so next time I can create a separate branch here and do my fixes there? If not please let me know how you expect me to work in here.

jkralik:
> Just to check what the intention with this is, is this so next time I can create a separate branch here and do my fixes there.
Precisely

PR #152, opened 2020-08-19
rename go-ocf to plgd-dev

jkralik:


cizmazia (approved):

jkralik:
There was legal issue with OCF (open connectivity foundation). We implemented spec https://openconnectivity.org/specs/OCF_Cloud_API_For_Cloud_Services_Specification_v2.2.0.pdf but they couldn't announce it as reference implemenatation, because they are not responsible for go-ocf. So they ask us to rename go-ocf(gocf). We had lot's of discussion with their layers but only acceptable way was the rename.

PR #174, opened 2020-10-26
Make DTLS Peer Certificates Available

### LINKED ISSUE #134: DTLS peer certificates
I noted that in v2.0.2 I can't get the peer certificates any more, which were added in #115. Am I just looking in the wrong place (`responseWriter.Client()`) or was that code removed somewhere in v2.0.1 ?

cristovao-trevisan:
Hello and thanks for the awesome work,

I'm building an IOT project, using COAP, and decided to use client certificates for auth (PKI).
As I understand, being able to read the client certificate for each connection is a must for everyone using PKI for auth.

As per #134, I added the option to set a context value (which can be the peer certificate), as well as a way to get read the certificate (using `WithOnNewClientConn`). I've done this only for DTLS and UDP (because UDP shares interfaces with DTLS), not for TCP or TLS (as I'm not using TCP).

Also added an example for PKI (with both server and client), as well as a markdown with instructions to create the certificates using openssl.

closes #134 

PS.: I understand that the example parses the certificate twice (once inside pion/dtls and other in the example code), and I'm not sure how expensive this operation is, so I just leave as it is.


cristovao-trevisan:
The *GitGuardian Security Checks* doesn't seem to like the certificate files used in the example and test. Any advice on how to fix this?

udp/session.go
```diff
@@ -41,6 +41,10 @@ func NewSession(
 	}
 }
 
+func (s *Session) SetContextValue(key interface{}, val interface{}) {
+	s.ctx = context.WithValue(s.ctx, key, val)
```
jkralik, line 45: Before `s.ctx` variable points to same address for whole time, with the new feature we need to guarded with lock.
eg `s.ctx` as atomic.Value

dtls/server.go
```diff
@@ -36,7 +37,7 @@ type GoPoolFunc = func(func()) error
 
 type BlockwiseFactoryFunc = func(getSendedRequest func(token message.Token) (blockwise.Message, bool)) *blockwise.BlockWise
 
-type OnNewClientConnFunc = func(cc *client.ClientConn)
+type OnNewClientConnFunc = func(cc *client.ClientConn, dtlsConn *dtls.Conn)
```
jkralik, line 40: Pls add comment that `dtlsConn *dtls.Conn` below to `client.ClientConn` and `dtlsConn.Close()` is forbidden -> cause panic for duplicity close.

jkralik:
Hi. Thanks for contributing. I want to add it but some changes i want to propose.

- _GitGuardian Security Checks_ it is better when you generate certificates than commit them to the repository.
- pls could extends tcp and tcp-tls because similar feature I need too :), I want to used it in https://github.com/plgd-dev/cloud/tree/master/coap-gateway

dtls/session.go
```diff
@@ -71,6 +71,10 @@ func (s *Session) Context() context.Context {
 	return s.ctx
 }
 
+func (s *Session) SetContextValue(key interface{}, val interface{}) {
```
jkralik, line 74: Before s.ctx variable points to same address for whole time, with the new feature we need to guarded with lock.
eg s.ctx as atomic.Value

cristovao-trevisan:
- I'm not familiar with `atomic.Value`, but all the session structs had a mutex, so I used it to guard it.
- Added similar functionality to TCP (and TLS), with test
- Removed the openssl generated certificates and keys in favor of golang generated ones (reason for the forced push)

cristovao-trevisan:
ping @jkralik

jkralik (commented):
I created new PR( #175 ) where I solved concurrent access to session.ctx. Thx for contributing - it is very nice feature !

cristovao-trevisan:
Thanks for everything. Any plans for when to make a new release?

jkralik:
Yes. I released [v2.1.0](https://github.com/plgd-dev/go-coap/releases/tag/v2.1.0) .

PR #189, opened 2020-12-25
Test that issue #157 is fixed

### LINKED ISSUE #157: ResponseWriter.SetResponse does not work with empty reader
This does lead to an EOF

    w.SetResponse(codes.Created, msg.AppCBOR, bytes.NewReader([]byte{}))

To produce an empty payload one has to call

    w.SetResponse(codes.Created, msg.AppCBOR, nil)

either the first case should just work, or the usage should be documented on the `SetResponse` method.

Niondir:
Fixed #157

The issue was fixed somewhere already. I added a Test that fails with Tag v2.0.4 but works now, to ensure it stays fixed in future.

### DIFF
--- a/udp/client_test.go
+++ b/udp/client_test.go
@@ -55,6 +55,15 @@ func TestClientConn_Get(t *testing.T) {
 			wantPayload:       []byte("b"),
 		},
 		{
+			name: "ok-empty",
+			args: args{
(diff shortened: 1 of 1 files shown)


jkralik (approved):
Thx! 👍

PR #190, opened 2020-12-26
Deduplicate messages with the same message ID for UDP

### LINKED ISSUE #156: Are ACKs for Confirmable requests and message deduplication handled by this library?
Hi,
I cant find any information about this.

I know the old version of this library did not automatically sent ACK responses. Also request deduplications based on messageIDs for retries had to be done in the app. Is this still required?

Niondir:
Fixes #156

As described in RFC 7252 4.5. Message Deduplication

- Handlers are only called once per received message id
- Responses are cached for EXCHANGE_LIFETIME = 247s (due to RFC 7252)
- Parallel handling of the same message ID is avoided to ensure propper caching
- Removed redundant setting of message ID and Type in func (cc *ClientConn) do(...)
- Fixed an error formatting mistake in connUDP.go

### DIFF
--- a/udp/client/clientconn.go
+++ b/udp/client/clientconn.go
@@ -8,6 +8,7 @@ import (
 	"sync/atomic"
 	"time"
 
+	"github.com/patrickmn/go-cache"
 	"github.com/plgd-dev/go-coap/v2/message"
 	"github.com/plgd-dev/go-coap/v2/net/blockwise"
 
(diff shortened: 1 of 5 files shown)


udp/client/clientconn.go
```diff
@@ -111,8 +117,7 @@ func (cc *ClientConn) do(req *pool.Message) (*pool.Message, error) {
 		return nil, fmt.Errorf("invalid token")
 	}
 
-	req.SetMessageID(cc.getMID())
-	req.SetType(udpMessage.Confirmable)
```
jkralik, line 115: These lines was mainly used for first request (GET/POST/...) and blockwise transfer. So if you remove them blockwise transfer sends NonConfirmable messages. So I want to suggest to move removing lines to next PR #191 and solve the issue there.

Niondir: Makes sense. I leave it to #191 where CON messages are getting the default for new messages.

jkralik (approved):

PR #192, opened 2020-12-27
Allow client to send non-confirmable messages

### LINKED ISSUE #191: Allow to send non confirmable requests via UDP
In `udp/client/clientconn.go` all messages are set to "Confirmable"

See:
```
func (cc *ClientConn) writeMessage(req *pool.Message) error {
	req.SetMessageID(cc.getMID())
	req.SetType(udpMessage.Confirmable)
```

I don't see the reason why. It removes the ability to send non confirmable messages via UDP.

When just removing the line `req.SetType(udpMessage.Confirmable)` at least some tests are failing. Seems like the client can't deal with not getting any response and should return `nil` rather than time out.

Niondir:
Fixes #191

- Non-confirmable requests are getting Non-confirmable responses by default
- Mux Message has IsConfirmable field for handler
- Blockwise transfer is keeping UDP message types
- ClientConn midHandlerContainer is only used for CON messages
- Non-confirmable requests are never re-transmitted
- The default type for new messages is Confirmable for backwards compatibility
- Added related testcases and asserts

### DIFF
--- a/net/blockwise/blockwise.go
+++ b/net/blockwise/blockwise.go
@@ -4,7 +4,6 @@ import (
 	"bytes"
 	"context"
 	"fmt"
-	udpPool "github.com/plgd-dev/go-coap/v2/udp/message/pool"
 	"io"
 	"sync"
 	"time"
@@ -15,6 +14,7 @@ import (
 	"github.com/patrickmn/go-cache"
 	"github.com/plgd-dev/go-coap/v2/message"
 	"github.com/plgd-dev/go-coap/v2/message/codes"
+	udpMessage "github.com/plgd-dev/go-coap/v2/udp/message"
 )
 
 // Block Opion value is represented: https://tools.ietf.org/html/rfc7959#section-2.2
@@ -114,6 +114,13 @@ type Message interface {
 	SetBody(r io.ReadSeeker)
 	SetSequence(uint64)
 	String() string
+}
+
+// hasType enables access to message.Type for supported messages
+// Since only UDP messages have a type
+type hasType interface {
+	Type() udpMessage.Type
+	SetType(t udpMessage.Type)
 }
 
 // EncodeBlockOption encodes block values to coap option.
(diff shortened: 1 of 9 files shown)


net/blockwise/blockwise.go
```diff
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"context"
 	"fmt"
+	udpPool "github.com/plgd-dev/go-coap/v2/udp/message/pool"
```
jkralik, line 7: This is ugly. It will be better when we move github.com/plgd-dev/go-coap/v2/udp/message/type.go to ithub.com/plgd-dev/go-coap/v2/message/types/type.go and extends the Message interface about SetType and Type functions (for tcp it do will be do nothing) and for UDP it will be proper used.
What do you think ?

Niondir: Was thinking the same. Maybe even a new interface just for the "Type" thing.
I just don't wanted to change the generic message package. I will update the PR to avoid the dependency to the UDP package here.

Niondir: I would move it to `github.com/plgd-dev/go-coap/v2/message/type.go`

Niondir: Any specific reasons for a separate `types` package? I would keep the number of packages lower and just put it to `.../go-coap/v2/message/type.go`

I can see that `codes` have their own package. Whats the motivations?

jkralik: `types` directly doesn't belongs to common message (`go-coap/v2/message`). Maybe you can leave it in `github.com/plgd-dev/go-coap/v2/udp/message` and import that package.

Sure I m ok with simple interface `Type`

Niondir: It's a question of what the expected scope of `go-coap/v2/message` is. 

I moved it back, but are using the interface now.

BTW thanks for invinting me to the project. I would still leave the actual merging to you.

jkralik: I m happy from with your contribution 👍

jkralik (approved):
Just resolve conflict and you can merget it. Thx

PR #193, opened 2020-12-28
Allow to set UDP ClientConn timing parameters

### LINKED ISSUE #180: Anyone opposed to a methos to adjust the retransmit timers and counters on a connection
Hello,

I have an IoT use case where the server is in the cloud, devices connect and I extract the client connection to talk back to the device.  I need to be able to tweak the retransmit timers and counters on that connection (based on data stored in my server).  Currently the values in the ClientConn are private, so I would like to expose them with a simple method...

Any objections?  I can create a pull request...

Niondir:
Fixes #180 

Adds SetTransmissionNStart, SetTransmissionAcknowledgeTimeout, SetTransmissionMaxRetransmit

It's not threadsafe, do we needs an additional Mutex and "safe" methods for using the values?

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -6,6 +6,7 @@ require (
 	github.com/pion/dtls/v2 v2.0.1-0.20200503085337-8e86b3a7d585
 	github.com/plgd-dev/kit v0.0.0-20200819113605-d5fcf3e94f63
 	github.com/stretchr/testify v1.5.1
+	go.uber.org/atomic v1.6.0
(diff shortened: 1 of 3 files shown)


jkralik:
Yes. It must be thread-safe becasue get/set can be used from different go routines. You can use `atomic.Value` instead of mutex for each value.

Niondir:
Are you okay with using `go.uber.org/atomic`. Maybe an alternative is to copy the wrappers over from there to remove the dependency.

jkralik (approved):

jkralik:
Thx

PR #194, opened 2020-12-28
Inactivity monitoring for UDP server

### LINKED ISSUE #178: Connection stays open when disabling keep-alive
I think the connections might stay open indefinitely in the DTLS listener when using it with `dtls.WithKeepAlive(nil)`. I expect that when disabling keep-alive, we at least would still see the CPU usage and number of goroutines drop back to zero/original after the context duration is expired or something, but after an hour it is still steady high.

We ran into a problem where our server, using `go-coap`, stays on a high CPU usage after clients have finished there connections. At that point there have been e.g. 500 connections, of which the clients are all finished now and don't sent any more messages, and no new incoming connections. I've been trying to debug the issue and found that when I remove `dtls.WithKeepAlive(nil)` from the DTLS server configuration `dtls.NewServer`, so enable keep-alive, the CPU drops back to zero as expected.

I've also seen that the goroutines count stays high and stable when disabling keep-alive and drop back to the original value with keep-alive enabled.

@jkralik Do you have any ideas whether this is indeed the case or maybe something else is going on? I'm looking into this further myself as well, but I'm not as familiar with the code of course.

Niondir:
Fixes #178

It's only for UDP yet, DTLS is missing yet.

I also updated the connection cleanup test to count go-routines to be sure everything is cleaned up.

### DIFF
--- a/udp/client/clientconn.go
+++ b/udp/client/clientconn.go
@@ -427,7 +427,7 @@ func (cc *ClientConn) Ping(ctx context.Context) error {
 	return fmt.Errorf("unexpected response(%v)", resp)
 }
 
-// Run reads and process requests from a connection, until the connection is not closed.
+// Run reads and process requests from a connection, until the connection is closed.
 func (cc *ClientConn) Run() error {
 	return cc.session.Run(cc)
 }
@@ -529,6 +529,7 @@ func (cc *ClientConn) getResponseFromCache(mid uint16, resp *pool.Message) (bool
 }
 
 func (cc *ClientConn) Process(datagram []byte) error {
+	cc.lastActivity = time.Now()
 	if cc.session.MaxMessageSize() >= 0 && len(datagram) > cc.session.MaxMessageSize() {
 		return fmt.Errorf("max message size(%v) was exceeded %v", cc.session.MaxMessageSize(), len(datagram))
 	}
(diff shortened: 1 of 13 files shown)


udp/client/clientconn.go
```diff
@@ -469,6 +470,7 @@ func (cc *ClientConn) Sequence() uint64 {
 }
 
 func (cc *ClientConn) Process(datagram []byte) error {
+	cc.lastActivity = time.Now()
```
jkralik, line 473: Pls use atomic.Value to store lastActivity and move it to inactivityMonitor struct.

jkralik:
Pls could you make it more generic (creating package net/monitor/inactivity) and use it in tcp/udp/dtls ?

Niondir:
I also added it for DTLS, does it make sense for TCP at all, here we have real connections?

I will move it to `net/inactivity` with the next commit.

jkralik:
I want to change keepalive implementation to use inactivity monitor - so ping will be fired when inactivity occurs.

Niondir:
I also saw some similarities to the keepalive implementation. Whats missing would be the inactivity monitoring for Clients and TCP which was not needed for this issue.
Also the ClientConn will need more calls to notify for activity. Or maybe even the Session could do the job on writeMessage.

I'm done with this PR now. Feel free to have a final look and merge it.

jkralik (approved):
Thx for contribution. 👍

Issue #11, opened 2018-10-12
Add support for send messages Release/Abort for CSM (TCP connection)


jkralik:
using CSM release/abort messages in clients and server.Shutdown (especially important with containers since you should have a SIGINT handler that is able to shutdown gracefully)

PR #199, opened 2021-01-25
tcp: handle signals in connection goroutine

jkralik:


tcp/session.go
```diff
@@ -175,37 +175,37 @@ func (s *Session) handleBlockwise(w *ResponseWriter, r *pool.Message) {
...
+		return true
 	case codes.Abort:
 		if r.HasOption(coapTCP.BadCSMOption) {
 			//TODO
 		}
-		return
+		return true
 	}
-	s.handleBlockwise(w, r)
+	return false
```
ondrejtomcik, line 208: What about single `default: return false` and here return true? It would be more readable.

tcp/session.go
```diff
@@ -175,37 +175,37 @@ func (s *Session) handleBlockwise(w *ResponseWriter, r *pool.Message) {
...
 			atomic.StoreUint32(&s.peerMaxMessageSize, size)
 		}
 		if r.HasOption(coapTCP.BlockWiseTransfer) {
 			atomic.StoreUint32(&s.peerBlockWiseTranferEnabled, 1)
 		}
-		return
+		return true
 	case codes.Ping:
 		if r.HasOption(coapTCP.Custody) {
 			//TODO
```
ondrejtomcik, line 193: What kind of TODO needs to be done here? Please create issue for that on github.

ondrejtomcik: https://github.com/plgd-dev/go-coap/issues/11

ondrejtomcik (approved):

PR #205, opened 2021-02-03
all protocols: refactore keepalive implementation

jkralik:
ping is send when connection is inactive some duration
and when it reach limit number of pings it calls inactive
function where user can close socket and print log error.

it doesn't use separate goroutine anymore for keepalive per session


dtls/options.go
```diff
@@ -104,20 +104,36 @@ func WithGoPool(goPool GoPoolFunc) GoPoolOpt {
 
 // KeepAliveOpt keepalive option.
 type KeepAliveOpt struct {
-	keepalive *keepalive.KeepAlive
+	limitFails uint32
+	duration   time.Duration
```
ondrejtomcik, line 108: Same as in kit, probably it belongs here:
new config for coapgw proposed.

```
      keepalive:
        enabled: true
        timeout: 5s
        maxRetries: 6
```

could you please already reflect it in this MR ?

jkralik: Done

ondrejtomcik (approved):

PR #208, opened 2021-02-18
Fix an issue which prevents blockwise xfer working with observe

Kegsay:
If you observe a resource and the state change is too big for a single
packet, it will be sent blockwise. However, only the first block is
sent with the observe token, it is expected that clients will request
subsequent blocks by issuing a new request. Returning a response to
this new request would fail on the client due to ETags not matching,
despite no ETags being used. The failure error was:

  received message doesn't contains ETAG but cached received message contains it([])

This patch ensures that the received message actually contains an
ETag, which fixes this issue.


jkralik:
Thx for the fix. Pls could you create a test for the future ?

jkralik (approved):

jkralik:
Ok. Thx I merge it.

Issue #214, opened 2021-03-09
fixing  example code on readme


itsmekingtiger:
on [example code's](https://github.com/plgd-dev/go-coap#server-udptcp) handler should be changed from

```go
func handleA(w mux.ResponseWriter, req *message.Message) {...}
```

to

```go
func handleA(w mux.ResponseWriter, req *mux.Message) {...}
```

is it?

PR #217, opened 2021-03-23
Fix example in README

mladedav:
As reported in #214 the example in README is wrong. This pull request fixes just that issue.


jkralik (approved):

mladedav:
Can this be merged now? I am not really sure what the state is after approving it but before merging.

PR #237, opened 2021-07-06
udp/server: allow discover device by unicast

jkralik:


udp/discover.go
```diff
@@ -23,18 +23,18 @@ type MulticastOption interface {
 }
 
 // Discover sends GET to multicast address and wait for responses until context timeouts or server shutdown.
```
rafajpet, line 25: multicast or unicast right ? I would suggest describe it more in docs comment

jkralik: Fixed

rafajpet (approved):

PR #243, opened 2021-07-20
Remove unused vars and improve code readability

arun1587:
hey, was able to make use of linters to fix some code there by improving the readability a bit. 
imo, these changes shouldn't have functional or performance impact.

### DIFF
--- a/message/option.go
+++ b/message/option.go
@@ -243,13 +243,13 @@ func extendOpt(opt int) (int, int) {
 func marshalOptionHeaderExt(buf []byte, opt, ext int) (int, error) {
 	switch opt {
 	case ExtendOptionByteCode:
-		if buf != nil && len(buf) > 0 {
+		if len(buf) > 0 {
 			buf[0] = byte(ext)
 			return 1, nil
 		}
 		return 1, ErrTooSmall
 	case ExtendOptionWordCode:
-		if buf != nil && len(buf) > 1 {
+		if len(buf) > 1 {
 			binary.BigEndian.PutUint16(buf, uint16(ext))
 			return 2, nil
 		}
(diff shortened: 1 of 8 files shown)


message/option.go
```diff
@@ -243,13 +243,13 @@ func extendOpt(opt int) (int, int) {
 func marshalOptionHeaderExt(buf []byte, opt, ext int) (int, error) {
 	switch opt {
 	case ExtendOptionByteCode:
-		if buf != nil && len(buf) > 0 {
```
arun1587, line 246: length of a nil buffer would be `0` so `buf !=nil` check is redundant.

message/options.go
```diff
@@ -558,7 +558,7 @@ func (options Options) Clone() (Options, error) {
 	opts, used, err := opts.ResetOptionsTo(buf, options)
 	if err == ErrTooSmall {
 		buf = append(buf, make([]byte, used-len(buf))...)
-		opts, used, err = opts.ResetOptionsTo(buf, options)
```
arun1587, line 561: `used` var isn't really used anywhere beyond this line. hence it can be omitted.

net/dtlslistener.go
```diff
@@ -162,13 +162,11 @@ func (l *DTLSListener) Accept() (net.Conn, error) {
 	}
 
 	if deadline.IsZero() {
-		select {
```
arun1587, line 165: use of select for a single case is unnecessary.

mux/middleware.go
```diff
@@ -5,19 +5,12 @@ package mux
...
-}
-
 // Middleware allows MiddlewareFunc to implement the middleware interface.
 func (mw MiddlewareFunc) Middleware(handler Handler) Handler {
 	return mw(handler)
 }
 
 // Use appends a MiddlewareFunc to the chain. Middleware can be used to intercept or otherwise modify requests and/or responses, and are executed in the order that they are applied to the Router.
 func (r *Router) Use(mwf ...MiddlewareFunc) {
-	for _, fn := range mwf {
```
arun1587, line 20: remove unnecessary for loop

jkralik (approved):
Great and thanks for contributing.

PR #247, opened 2021-09-08
udp/tcp/dtls: Done() signalizes that connection is not more processed.

jkralik:


der-dano (commented):
Shouldn't we add a simple test case for this new flag?

jkralik:
> Shouldn't we add a simple test case for this new flag?

Done

der-dano (approved):

PR #257, opened 2021-11-16
udp/server: support piggyback response

ManuelBk:
According to RFC7252 5.2.1 Responses can
be sent piggybacked to the Acknowledgement.
The note states, that "there is a strong
expectation that servers will implement
code to piggyback whenever possible".

### DIFF
--- a/udp/client/clientconn.go
+++ b/udp/client/clientconn.go
@@ -663,8 +663,8 @@ func (cc *ClientConn) Process(datagram []byte) error {
 				return
 			}
 			return
-		} else if reqType == udpMessage.Confirmable {
-			// send separate message to confirm received message.
+		} else if reqType == udpMessage.Confirmable && !w.response.IsModified() {
+			// send message to confirm received message, if response is not modified
 			separateMessage := cc.AcquireMessage(cc.Context())
 			defer cc.ReleaseMessage(separateMessage)
 			separateMessage.SetCode(codes.Empty)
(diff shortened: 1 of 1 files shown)


udp/client/clientconn.go
```diff
@@ -672,9 +672,9 @@ func (cc *ClientConn) Process(datagram []byte) error {
 			return
 		}
 
-		// send message with confirmation
-		w.response.SetType(udpMessage.Confirmable)
-		w.response.SetMessageID(cc.getMID())
+		// send piggybacked response
+		w.response.SetType(udpMessage.Acknowledgement)
```
jkralik, line 676: Hi. We need to handle when we send `udpMessage.Acknowledgement`. With this response, it always sends `udpMessage.Acknowledgement` but for blockwise transfer, we want to send a confirmation message.
So change this line to something like:
```
w.response.SetType(udpMessage.Confirmable)
if reqType == udpMessage.Confirmable {
  w.response.SetType(udpMessage.Acknowledgement)
} ;
```

ManuelBk: Hi, thanks for the review and sorry for my long absence. I have incorporated your request and hope it is fine now :)

udp/client/clientconn.go
```diff
@@ -682,9 +682,12 @@ func (cc *ClientConn) Process(datagram []byte) error {
...
 		}
 
-		// send message with confirmation
+		// send piggybacked response
 		w.response.SetType(udpMessage.Confirmable)
-		w.response.SetMessageID(cc.getMID())
+		if reqType == udpMessage.Confirmable {
+			w.response.SetType(udpMessage.Acknowledgement)
+		}
+		w.response.SetMessageID(reqMid)
```
jkralik, line 690: when you `reqType !=  udpMessage.Confirmable` you need to call `w.response.SetMessageID(cc.getMID())` instead of w.response.SetMessageID(cc.getMID()) . otherwise you use same messageId as the peer send and peer can drop the message when he received the response (deduplication).

jkralik (approved):

jkralik:
@ManuelBk Thx for your contribution!

ManuelBk:
> @ManuelBk Thx for your contribution!

It was a pleasure, thanks for reviewing!

PR #260, opened 2021-11-26
optimizations of caches and network stack

jkralik:
- net: remove set deadlines - goroutines are not wakeup for timeouts
- introduce better cache for expirations


dtls/server.go
```diff
@@ -252,14 +292,20 @@ func (s *Server) Serve(l Listener) error {
 // Stop stops server without wait of ends Serve function.
 func (s *Server) Stop() {
 	s.cancel()
+	s.listenMutex.Lock()
+	l := s.listen
```
Danielius1922, line 296: This seems a bit weird, why bother with the temp variable `l` when you keep holding the mutex anyway?

Perhaps it was means to be:
```
s.listenMutex.Lock()
l := s.listen
s.listen = nil
s.listenMutex.Unlock()
if l != nil {
  l.Close()
}
```

Which avoid calling Close while holding the mutex. But if that's not a problem then `l` can be removed.

net/connUDP.go
```diff
@@ -16,13 +17,11 @@ import (
...
-	packetConn     packetConn
-	errors         func(err error)
-	network        string
-	onReadTimeout  func() error
-	onWriteTimeout func() error
+	connection *net.UDPConn
+	packetConn packetConn
+	errors     func(err error)
+	network    string
+	closed     uint32
```
Danielius1922, line 24: Using `atomic.Bool` would be safer, somebody could use non-atomic operations with `uint32` by mistake. Non-atomic comparisons are disabled for `atomic` variables.

tcp/server_test.go
```diff
@@ -253,20 +262,24 @@ func TestServer_KeepAliveMonitor(t *testing.T) {
...
+
 	require.NoError(t, err)
 	checkCloseWg.Add(1)
 	cc.AddOnClose(func() {
 		checkCloseWg.Done()
 	})
 
 	// send ping to create serverside connection
-	ctx, cancel := context.WithTimeout(context.Background(), time.Second)
+	ctx, cancel = context.WithTimeout(context.Background(), time.Second)
```
Danielius1922, line 279: Doesn't this overwrite the defered `cancel` on line 232? (I always forget how exactly does defer behave :zipper_mouth_face: )

udp/client/clientconn.go
```diff
@@ -8,18 +8,19 @@ import (
...
+	"github.com/plgd-dev/go-coap/v2/net/blockwise"
+	"github.com/plgd-dev/go-coap/v2/net/monitor/inactivity"
+	"github.com/plgd-dev/go-coap/v2/pkg/cache"
 	udpMessage "github.com/plgd-dev/go-coap/v2/udp/message"
 	"github.com/plgd-dev/go-coap/v2/udp/message/pool"
 	kitSync "github.com/plgd-dev/kit/v2/sync"
+	atomicTypes "go.uber.org/atomic"
 )
 
+const ExchangeLifetime = 247 * time.Second
```
Danielius1922, line 22: 247? It that some OCF constant?

Danielius1922 (approved):

PR #264, opened 2021-12-05
efficient struct packing

jkralik:


Danielius1922:
I still get one report of suboptimal struct size:

```
$ aligncheck ./...
github.com/plgd-dev/go-coap/v2/message/pool: /home/danielius/go/src/github.com/plgd-dev/go-coap/message/pool/message.go:16:6: struct Message could have size 160 (currently 168)
```

message/getToken.go
```diff
@@ -11,6 +12,10 @@ func (t Token) String() string {
 	return hex.EncodeToString(t)
 }
 
+func (t Token) Hash() uint64 {
```
Danielius1922, line 15: We could cache the value if the function is used often, eg:

```
type Token struct {
  data []byte
  hash uint64
}
```

Calculate the value on first use, save it and then use only cached value.

jkralik: Hash is a very cheap operation so was can leave it as is otherwise we need to introduce a `constructor` which fills the value.

Danielius1922 (approved):

PR #265, opened 2021-12-05
introduce static analysis

jkralik:


JosefWN:
Tip, you can manage a whole host of linters easily and efficiently using: https://github.com/golangci/golangci-lint

Danielius1922:
> Tip, you can manage a whole host of linters easily and efficiently using: https://github.com/golangci/golangci-lint

Indeed, thanks, we have it included in the plgd-dev/hub repository. I've added there both stand-alone analysis and golangsci-lint, though I no longer recall why. (The latest released version and the version of the linter bundled in golangsci might be different? Can't remember, lets just say that I'm a linting maximalist.)

message/pool/message.go
```diff
@@ -75,6 +76,9 @@ func (r *Message) ResetOptionsTo(in message.Options) {
 		r.valueBuffer = append(r.valueBuffer, make([]byte, used)...)
 		opts, used, err = r.msg.Options.ResetOptionsTo(r.valueBuffer, in)
 	}
+	if err != nil {
+		panic(fmt.Errorf("unexpected error: %w", err))
```
Danielius1922, line 80: @jkralik panicking in these setters feels wrong. But what else can we do without changing the API? If the first call `r.msg.Options.ResetOptionsTo` gives us correct required size for the value buffer this should never occur. But any ideas how else might we enforce it?

Danielius1922 (approved):

PR #269, opened 2021-12-17
Fix issues reported by errcheck

Danielius1922:


jkralik (approved):

tcp/clientobserve.go
```diff
@@ -79,7 +79,9 @@ func (o *Observation) handler(w *ResponseWriter, r *pool.Message) {
 }
 
 func (o *Observation) cleanUp() bool {
-	o.cc.observationTokenHandler.Pop(o.token)
+	if _, err := o.cc.observationTokenHandler.Pop(o.token); err != nil {
```
jkralik, line 82: ignore error - it means that observation does not exist at all.

Danielius1922: `o.cc.observationRequests.PullOut` should still be called in that case?

Danielius1922: Updated

PR #277, opened 2022-01-11
Add multicast and non-confirmed message support

### LINKED ISSUE #272: NonConfirmable message gets always turned into Confirmable block messages
When experimenting with outgoing client multicast (NonConfirmable, NoResponse-2) I noticed that the function WriteMessage() in udp/client/clientconn.go automatically converts all messages into Confirmable (block-wise, even with a single block) messages.

I am not sure whats the best fix for this... 
* turn off blockwise transfer for NonConfirmable messages ?
* copy message type to generated blockwise messages ?

HRogge:
This patch fixes #272 and adds single-interface multicast capability to go-coap.

I pushed the unified multicast options (used by the new send-multicast function and the existing discover function) into a package of their own because they are used by two different packages and I got cyclical import problems first. I am not thrilled about the solution, so I would happily move the options somewhere else.


HRogge:
Hmm... I think some of these errors were already in the existing code... could it be the linter got better? Shall I try to fix them all?

net/connUDP.go
```diff
@@ -208,86 +207,64 @@ func (c *UDPConn) Close() error {
...
-		return fmt.Errorf("cannot write multicast with context: cannot get interfaces for multicast connection: %w", err)
-	}
-	c.lock.Lock()
-	defer c.lock.Unlock()
-	for _, iface := range ifaces {
-		if iface.Flags&net.FlagMulticast == 0 {
-			continue
-		}
-		ifaceAddrs, err := iface.Addrs()
+	if iface == nil {
```
jkralik, line 248: pls, add an option set that allows sending the message without a defined interface. (it uses any interface chosen by OS)

https://github.com/plgd-dev/go-coap/blob/jkralik/feature/multicastToAnyInterface/udp/discover.go#L107

HRogge: you can just use "nil" for the interface and it automatically cycles through all available interfaces (like the code did before).
I am not sure the concept of "what the operation system chooses" really works well, you often end up using a unicast default route.

But it wouldn't be difficult to add a "WithAnyInterface()" multicast option... just needs an additional flag in the mcoption struct to differentiate between "all interfaces" and "any interface".

jkralik: `WithAnyInterface()` is needed for android 30+.  The call `net.Interfaces()` returns an error with permission denied so I want to have the option to use any interface.

HRogge: I have pushed an update, please see if this works for you (especially from the API interface).

Danielius1922:
> Hmm... I think some of these errors were already in the existing code... could it be the linter got better? Shall I try to fix them all?

Just check if there's a problem in your newly added code, ignore the rest. We've just enabled several new linters, I'm working on cleaning it up, but it requires more work than I expected. Thanks.

HRogge:
I think the third commit should fix the existing errors in the changed file.

udp/client/clientconn.go
```diff
@@ -159,18 +161,22 @@ func (cc *ClientConn) do(req *pool.Message) (*pool.Message, error) {
...
+		})
+		if err != nil {
+			return nil, fmt.Errorf("cannot add token handler: %w", err)
 		}
-	})
-	if err != nil {
-		return nil, fmt.Errorf("cannot add token handler: %w", err)
+		defer func() { _, _ = cc.tokenHandlerContainer.Pop(token) }()
+	} else {
+		close(respChan)
```
jkralik, line 177: The purpose of `do` is to wait for the response message (eg NonConfirmable). With `close(respChan)` message never comes.  Why did you change it?

udp/client/clientconn.go
```diff
@@ -195,8 +201,10 @@ func (cc *ClientConn) Do(req *pool.Message) (*pool.Message, error) {
 		req.UpsertMessageID(cc.getMID())
 		return cc.do(req)
 	}
+	msgType := req.Type()
 	bwresp, err := cc.blockWise.Do(req, cc.blockwiseSZX, cc.session.MaxMessageSize(), func(bwreq blockwise.Message) (blockwise.Message, error) {
 		req := bwreq.(*pool.Message)
+		req.SetType(msgType)
```
jkralik, line 207: At blockwise it is recommended to use `.Confirmable` messages:
```
   specification minimizes the constraints it adds to those base
   exchanges; however, not all variants of using CoAP are very useful
   inside a block-wise transfer (e.g., using Non-confirmable requests
   within block-wise transfers outside the use case of Section 2.8 would
   escalate the overall non-delivery probability).
   ```
We need to change update `get,post,delete,put` operations to use `Confirmable` type by default.

File: udp/client/clientconn.go (hunk 2 above)
Comment (HRogge) [reply udp/client/clientconn.go:177]: A non-confirmable request doesn't have a response... so there is no point in waiting for it.

Comment (jkralik) [reply udp/client/clientconn.go:177]: This is not true: https://datatracker.ietf.org/doc/html/rfc7252#section-5.2.3

Comment (HRogge) [reply udp/client/clientconn.go:177]: Hmm... okay, now I am a bit confused... I am using Non-Confirmable messages in terms of Multicast, which (normally?) has no response.

If non-confirmable messages can have a response, how does the client KNOW if it gets a response or not? Or do we need a "WithNoResponseExpected()" Option?

Comment (jkralik) [reply udp/client/clientconn.go:177]: There is a way to set a deadline in context so that if the response does not arrive then the error deadline exceeded is returned.

**Comment by jkralik** [comment]
I want to suggest removing non-confirmed message support to separate PR. Because I think it is more complex.

File: udp/discover.go [hunk 4]
```diff
@@ -40,14 +28,14 @@ func (s *Server) Discover(ctx context.Context, address, path string, receiverFun
 // DiscoveryRequest sends request to multicast/unicast address and wait for responses until request timeouts or server shutdown.
 // For unicast there is a difference against the Dial. The Dial is connection-oriented and it means that, if you send a request to an address, the peer must send the response from the same
 // address where was request sent. For Discover it allows the client to send a response from another address where was request send.
-func (s *Server) DiscoveryRequest(req *pool.Message, address string, receiverFunc func(cc *client.ClientConn, resp *pool.Message), opts ...MulticastOption) error {
+func (s *Server) DiscoveryRequest(req *pool.Message, address string, receiverFunc func(cc *client.ClientConn, resp *pool.Message), opts ...generic.MulticastOption) error {
```
jkralik, line 31: Please could you extend the test at https://github.com/plgd-dev/go-coap/blob/f28bb10a28501f0449d71675f5c6077a4ecaf657/udp/server_test.go#L79 about `generic.MulticastOption` ?

udp/client/clientconn.go
HRogge: I think setting a deadline is a bad choice... because the timing could be hardware (and cpu/network load) dependent. There should be an option that says "I don't care, just send it and come back".

HRogge:
As long as the WriteMulticast() call does not wait for a response I am okay with removing the NonConfirmable code change. I will also see what I can do about writing a few tests...

HRogge:
Didn't you say you removed (or wanted to remove) the "non-conformation" parts of the changes? I am looking over your merged branch now and there are some issues because (parts?) of the code is still inside.

PR #283, opened 2022-01-12
Fix the case when message in receive cache is nil

arun1587:


jkralik:
pls, could you fix the merge conflict? thx

jkralik:
Hi. I looked at PR and it seems that the fix is not needed anymore for the current master branch. Can I close the PR?

arun1587:
> @arun1587 Hi. I looked at PR and it seems that the fix is not needed anymore for the current master branch. Can I close the PR?

Yes please

PR #284, opened 2022-01-12
Update github workflows triggers

Danielius1922:


.github/workflows/test.yml
```diff
@@ -5,9 +5,9 @@ name: Test
 # Controls when the action will run.
 on:
   push:
-  pull_request:
     branches:
       - master
```
jkralik, line 9: is thi correct ?

Danielius1922: `pull_requests:`

Triggers for all PR, even from forked repo, push doesn't seem to. 

But push seems to be needed for `codecov` coverage, so I also need it, but it's enough to run when the PR is being merged to master. 

However, I don't see a comment from the `codecov` bot with coverage stats of this PR. :/ using only `push:` would run everything twice, which I don't like. I guess we need a separate workflow for the coverage stats of the PR.

``

Danielius1922: Created separate workflow for PR from forked repos, works for the use cases we support.

jkralik (approved):

PR #285, opened 2022-01-12
Fix issues reported by errcheck

Danielius1922:


server.go
```diff
@@ -22,16 +23,24 @@ func ListenAndServe(network string, addr string, handler mux.Handler) error {
 		if err != nil {
 			return err
 		}
-		defer l.Close()
 		s := udp.NewServer(udp.WithMux(handler))
+		defer func() {
+			if err := l.Close(); err != nil {
```
Danielius1922, line 28: Wasn't sure how to handle this, so there are three suggestions in this file

1) add public `Error` method to tcp/udp.Server that handles errors
2) use `log.Printf`
3) ignore it

@jkralik what do you think, which way is the best?

net/monitor/inactivity/inactivitymonitor.go
```diff
@@ -36,8 +35,8 @@ func (m *inactivityMonitor) LastActivity() time.Time {
 	return time.Time{}
 }
 
-func CloseClientConn(cc ClientConn) {
-	cc.Close()
+func CloseClientConn(cc ClientConn) error {
```
Danielius1922, line 38: API change, but adding `error` output shouldn't break most code (only cases I can think of is assigning to function pointer with previous signature or adhering to interface).

udp/message/getmid.go
```diff
@@ -15,6 +16,9 @@ func GetMID() uint16 {
 
 func RandMID() uint16 {
 	b := make([]byte, 4)
-	rand.Read(b)
+	_, err := rand.Read(b)
+	if err != nil {
+		log.Fatalf("cannot get random value: %v", err)
```
Danielius1922, line 21: Is there anything else we can do here? Aborting within a library should be avoided, but I couldn't think of anything else. It could return error, but the error would just move to

```go
var msgID = uint32(RandMID())
```
.

server.go
jkralik: you can handle this error at line 32. If `s.Serve` returns nil then you use error from `l.Close`

udp/message/getmid.go
jkralik: I think we can use the time when `rand.Read(b)` fails. It's just mid which is used for pairing messages. And this func i think is used only for initialization.,

net/monitor/inactivity/inactivitymonitor.go
jkralik: This is concrete default implemenetation. So for default, we can ignore the error.

Danielius1922: Actually, `func CloseClientConn` is not used anywhere. Do we need it at all? The caller can call `Close` on `CloseClientConn` directly without this wrapper.

server.go
Danielius1922: I guess that's alright, we can ignore the `Close` error if the `Serve` call fails.

jkralik (approved):

PR #289, opened 2022-01-13
tests: replace waitgroup by semaphore

jkralik:


dtls/client_test.go
```diff
@@ -712,13 +713,13 @@ func TestClientInactiveMonitor(t *testing.T) {
 		require.NoError(t, err)
 	}()
 
-	var checkCloseWg sync.WaitGroup
-	defer checkCloseWg.Wait()
+	checkCloseWg := semaphore.NewWeighted(2)
```
Danielius1922, line 716: checkCloseWg -> It's not a `waitGroup` anymore, so lets get rid of the suffix.

Danielius1922 (approved):

PR #291, opened 2022-01-14
Extends multicast options

jkralik:
* Add option to send multicast to an unspecified interface

Co-authored-by: Jozef Kralik <jojo.lwin@gmail.com>

### DIFF
--- a/net/connUDP.go
+++ b/net/connUDP.go
@@ -5,7 +5,6 @@ import (
 	"fmt"
 	"net"
 	"strings"
-	"sync"
 	"time"
 
 	"go.uber.org/atomic"
@@ -23,8 +22,6 @@ type UDPConn struct {
 	connection *net.UDPConn
 	errors     func(err error)
 	closed     atomic.Bool
-
-	lock sync.Mutex
 }
 
 type ControlMessage struct {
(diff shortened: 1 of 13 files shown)


net/connUDP.go
```diff
@@ -208,88 +205,173 @@ func (c *UDPConn) Close() error {
...
+			continue
+		}
+		filtered = append(filtered, srcAddr)
 	}
-	if err := p.SetMulticastHopLimit(multicastHopLimit); err != nil {
+	return filtered
+
+}
+
+func ConvAddrsToIps(ifaceAddrs []net.Addr) []net.IP {
```
Danielius1922, line 263: Do `ConvAddrsToIps` and `FilterAddressesByNetwork` need to be exported? Seems that they are utility functions only used in this file. We can keep them exported, but we should then add documentation and tests for them.

jkralik: Fixed

net/connUDP.go
```diff
@@ -208,88 +205,173 @@ func (c *UDPConn) Close() error {
...
+		addr := strings.Split(addrMask, "/")[0]
+		ip := net.ParseIP(addr)
+		if ip != nil {
+			ips = append(ips, ip)
+		}
+	}
+	return ips
+}
+
+func (c *UDPConn) WriteMulticast(ctx context.Context, raddr *net.UDPAddr, buffer []byte, opts ...MulticastOption) error {
```
Danielius1922, line 276: Add docu to public API, doesn't have to be long.

net/connUDP.go
```diff
@@ -208,88 +205,173 @@ func (c *UDPConn) Close() error {
...
+		return nil
+	}
+	if len(errors) == 1 {
+		return errors[0]
+	}
+	return fmt.Errorf("%v", errors)
 }
 
-func (c *UDPConn) WriteMulticast(ctx context.Context, raddr *net.UDPAddr, hopLimit int, buffer []byte) error {
+func (c *UDPConn) writeMulticast(ctx context.Context, raddr *net.UDPAddr, buffer []byte, opt MulticastOptions) error {
```
Danielius1922, line 318: SonarCloud complains that this method is too complex: 

> Refactor this method to reduce its Cognitive Complexity from 27 to the 15 allowed.Why is this an issue?

HRogge:
The changes look fine... 
I am still a bit unsure about the whole "iterate over all IPs of an interface because as far as I know the "source address" can be the unspecified address when you select an interface... is this a problem on Windows or Mac?

jkralik:
> The changes look fine... I am still a bit unsure about the whole "iterate over all IPs of an interface because as far as I know the "source address" can be the unspecified address when you select an interface... is this a problem on Windows or Mac?

There is a problem with overall platforms. For example, if you have multiple IP addresses on the interface, some listeners will have incorrect source IP addresses.

udp/client/clientconn.go
```diff
@@ -592,6 +594,66 @@ func (cc *ClientConn) CheckMyMessageID(req *pool.Message) {
...
+	defer cc.inactivityMonitor.Notify()
+	reqMid := req.MessageID()
+
+	// The same message ID can not be handled concurrently
+	// for deduplication to work
+	l := cc.msgIdMutex.Lock(reqMid)
+	defer l.Unlock()
+
+	if ok, err := cc.getResponseFromCache(req.MessageID(), w.response); ok {
+		fmt.Printf("ClientConn.Process.processReq.AAA: req %v resp %v\n", req, w.response)
```
Danielius1922, line 607: is this OK for production build?

udp/client/clientconn.go
```diff
@@ -613,113 +673,31 @@ func (cc *ClientConn) Process(datagram []byte) error {
...
-				closeConnection()
-				cc.errors(fmt.Errorf("cannot write response: %w", err))
-				return
+				cc.ReleaseMessage(req)
 			}
-			return
-		} else if err != nil {
+		}()
+		resp := cc.AcquireMessage(cc.Context())
+		resp.SetToken(req.Token())
```
Danielius1922, line 682: Previously
```
origResp.SetType(req.Type())
```

was called after `SetToken`, but the call to `SetType` is no longer here. I see that `cc.processReq` calls `w.response.SetType` in almost all code branches except for inner `if` for the case `w.response.IsModified() && (w.response.Type() == udpMessage.Reset || w.response.Code() == codes.Empty)`. So for that code path the response type might be different, is that OK?

Danielius1922: Also originally for `cc.getResponseFromCache(req.MessageID(), w.response); ok` with `ok = true` the code was 

```go
if req.Type() == udpMessage.Confirmable {
  w.response.SetType(udpMessage.Acknowledgement)
  w.response.SetMessageID(reqMid)
} else {
  w.response.SetType(udpMessage.NonConfirmable)
  w.response.SetMessageID(cc.getMID())
}
```

The `SetType` call is now missing in the else branch, is that OK?

net/connUDP.go
jkralik: Fixed

udp/client/clientconn.go
jkralik: Removed

HRogge:
I just tested the current version of the branch with my multicast test-code and it worked fine (after fixing the reference to the WithMulticastInterface function in my code). I will later test it in a network-namespace to make sure the interface selection works fine (which has to be done in the absence of a default-route to prevent Linux from faking things ^^)

HRogge:
I get the following error on a system without a default route:
```dial udp 224.0.1.187:5683: connect: network is unreachable```

I am calling ```coapudp.Dial("224.0.1.187:5683", coapudp.WithBlockwise(false, 0, 0))``` from my code.

I am not sure the code should call "connect()" for multicast targets, but I am still looking for the place in the code where it happens.

net/connUDP.go
jkralik: Fixed

HRogge:
I traced the origin of the error, its happening in
net/sock_posix.go:71
```	if err := fd.dial(ctx, laddr, raddr, ctrlFn); err != nil { ```

the problem at this point is that the API tries to build an UDP "connection", which will fail for a multicast address in the absence of a default route. The default route is not necessary for sending, but something in the setup go wrong.

We need a socket which has NOT set the remote address so that the socket API doesn't call "connect()" in sock_posix.go:150
connect() is mostly for making unicast connections easier.

HRogge:
Okay, I got the issue resolved with the help of a new option I call WithMulticastTarget(), which allows you to overwrite the destination of a WriteMulticast() call... this way you can open the socket without the multicast IP (e.g. ":5634" instead of "224.0.1.187:5683") but still can specify the destination of the Multicast.

How do you want to get the additions (a bit of code in net/connUDP.go and net/options.go)?

HRogge:
I can also provide the scripts I use to setup my network namespaces for testing...

udp/client/clientconn.go
jkralik: Fixed

jkralik:
> Okay, I got the issue resolved with the help of a new option I call WithMulticastTarget(), which allows you to overwrite the destination of a WriteMulticast() call... this way you can open the socket without the multicast IP (e.g. ":5634" instead of "224.0.1.187:5683") but still can specify the destination of the Multicast.
> 
> How do you want to get the additions (a bit of code in net/connUDP.go and net/options.go)?

You can use `(s *Server) DiscoveryRequest()` where you can put to server listen socket(":5634"). And then just call discovery. Im not sure with (s *Client) if this will be worked, because it doesn't support multiple peer sessions - just one.

Danielius1922 (changes requested):
SonarCloud complains about some smell again, otherwise I see nothing else.

HRogge:
I have (in my local repository) added an option to allow a client to send Multicast to several addresses by selecting a target for each multicast simultaneously.

Danielius1922 (approved):

HRogge:
Can we see if you add #293 to this change set before merging it to master? Multicast support for the server is not that useful if the client side doesn't work correctly...

Danielius1922 (approved):

jkralik:
> Can we see if you add #293 to this change set before merging it to master? Multicast support for the server is not that useful if the client side doesn't work correctly...

Please, could you explain your use case? 
https://github.com/plgd-dev/go-coap/pull/293/commits/3fcddc77af2a2e2c543ee82e01fa56b7a7e1f99e

Because as I said: If you want to discover clients you need to use 
https://github.com/plgd-dev/go-coap/blob/master/udp/discover.go#L29 or 
func (s *Server) DiscoveryRequest(req *pool.Message, address string, receiverFunc func(cc *client.ClientConn, resp *pool.Message), opts ...MulticastOption) error 

In the handler of responses, you will have the client which you can store somewhere for eg:
https://github.com/plgd-dev/go-coap/blob/master/examples/observe/server/main.go#L75

HRogge:
I am using Discovery/Hello-Messages in terms of networking protocols, not specifically coap. I am working for a German Fraunhofer Institute (Fraunhofer FKIE in Wachtberg) and we are doing work on decentral radio networks, so protocols with "Hello/Beacon messages" are pretty common. Messages which already contain information for the receivers and don't need any answer.

I am not sure if the COAP discovery mechanism is that useful for this purpose.

Originally all of this stuff was done in custom binary messages but I like the idea of having COAP in between and being able to use its semantics without having to reinvent the wheel.

My current project is a lightweight decentralized data store which use broad/multicasted Hash-values to communicate its current state (so that the "steady state" is low-traffic). Similar to the HNCP Homenet protocol (see RFC 7788) but useful for pure low/medium datarate radio networks, not for ethernet/wifi.

HRogge:
Just as an additional explanation, in slow radio networks "media access" is often more important than "data rate"... when using a server discovery via multicast (and redo it from time to time because of limited radio range) and then use unicast to get the data, you square the number of packets involved in the process, which can be really bad.

jkralik:
Ok, Now I understand :), So I added address as an argument in `WriteMulticastMessage in client` - it is much clear. Is it ok now ?

HRogge:
Just pulled your commit, changed a few lines in my code and it seems to work fine.

Network Namespaces are just fun for testing code...

HRogge:
hmm... just a thought... did anyone already test the code with IPv6 linklocal addresses and multicast?

Danielius1922 (approved):

PR #293, opened 2022-01-17
Add support for client to send multicast to specific address/port

HRogge:
Additional code parts to allow clients to specify multicast target per transmission to allow client multicast work in the absence of unicast default route.


jkralik:
Thx for your contribution.

HRogge:
You are welcome... it was a quite happy surprise to see my suggestions being merged THAT quickly.

If I find something interesting missing from the API while using it, I will come back with a suggesting/pull request.

Just one thing, do you need a copy of my network-namespace script I use to test the interface specific multicast? Its just three machines with point-2-point veth's between each other, one veth on each to the main system but no default routes.

jkralik:
@HRogge  I permitted you to go-coap so you can now directly contribute changes to go-coap.

PR #295, opened 2022-01-21
Add support for route params

Danielius1922:


message/options.go
```diff
@@ -50,15 +50,14 @@ func (options Options) path(buf []byte) (int, error) {
 		needed += len(options[i].Value)
 		needed++
 	}
-	needed--
+
 	if len(buf) < needed {
 		return needed, ErrTooSmall
 	}
 	for i := firstIdx; i < lastIdx; i++ {
-		if i != firstIdx {
```
Danielius1922, line 58: @jkralik:

I made @DavidMarquezF PR almost work, the problem is that this now always returns path starting with "/", but previously the "/" was always stripped. 

The failing test `TestPathOption` calls `SetPath` with "a/b/c" path. `SetPath` always strips the leading '/' so there is no way to distinguish between absolute or relative paths. If paths should be always absolute then there's no problem and I can simply modify the test case. If paths can be relative then there's a bigger problem, because the router matcher uses regexps that always start with "^/" and thus we need a bigger rework.

jkralik: By COAP RFC - path "/a/b/c"  is changed to array of UriPath option - something like [ "a", "b", "c" ] and it is same for path "a/b/c" -  [ "a", "b", "c" ] , so relative and absolute path are same.

jkralik (approved):

PR #296, opened 2022-01-25
message: add tests to increase code coverage

Danielius1922:


message/pool/message_test.go
```diff
@@ -0,0 +1,30 @@
...
+func TestMessageSetPath(t *testing.T) {
+	msg := NewMessage()
+	size := 4
+	for i := 0; i < 8; i++ {
+		buffer := make([]byte, size)
+		buffer[0] = '/'
+		_, err := rand.Read(buffer[1:])
+		require.NoError(t, err)
+		inPath := string(buffer)
+		msg.SetPath(inPath)
```
Danielius1922, line 24: SetPath crashes for paths longer than 255 chars,

[https://github.com/plgd-dev/go-coap/blob/c3e8d8de2f5c3a0cd69a2811200308613e6243bc/message/pool/message.go#L94-L99]

jkralik: if `used < 0` then the error `message.ErrTooSmall` is incorrect. We need to introduce error `message.ErrTooLong` for this case.

message/pool/message.go
```diff
@@ -90,19 +92,39 @@ func (r *Message) Options() message.Options {
...
+// (URI-Path segments must have maximal length of 255) the internal
+// buffer is expanded.
+// If the path is too long, but not valid then the function returns
+// ErrInvalidValueLength error.
+func (r *Message) SetPath(p string) error {
 	opts, used, err := r.msg.Options.SetPath(r.valueBuffer, p)
-
-	if err == message.ErrTooSmall {
-		r.valueBuffer = append(r.valueBuffer, make([]byte, used)...)
+	for err == message.ErrTooSmall {
```
Danielius1922, line 105: NitPick: I could use something like

```go
expandBy := len(p) - len(r.valueBuffer)
```

and avoid the forcycle.

message/pool/message_test.go
Danielius1922: Should be fixed now. 

I've also found an edge case when the path had duplicated '/' characters. I've fixed that too and added tests.

jkralik (approved):

PR #299, opened 2022-01-29
Fixes for SonarCloud

Danielius1922:


message/pool/message_test.go
```diff
@@ -140,3 +141,122 @@ func TestMessageMustSetPath(t *testing.T) {
...
+				value: []byte(strings.Repeat("a", maxETagPathLen+1)),
+			},
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			msg := pool.NewMessage()
+			msg.SetETag(tt.args.value)
+
+			value, err := msg.ETag()
```
Danielius1922, line 253: @jkralik we have both `GetETag` and `ETag` functions, I think that's by mistake right?

Also how strict should we take the option definition? It defines min length as 1 and max length as 8, but the setter doesn't have problem with values outside of this range.

jkralik: pls remove `GetETag`

I think we can be strict.

Danielius1922: `GetETag` removed.

I've added bounds check to `SetETag`, though technically that changes public API. But IMO adding `error` return value shouldn't be problematic.

message/pool/message.go
```diff
@@ -135,11 +135,20 @@ func (r *Message) SetCode(code codes.Code) {
 	r.isModified = true
 }
 
-func (r *Message) SetETag(value []byte) {
+// SetETag insert/update ETag option.
+//
+// Option definition:
+// 	- format: opaque, length: 1-8, repeatable
```
Danielius1922, line 141: @jkralik apparently, ETag should be repeatable. I guess we should add a `AddETag` function and the setter should remove all existing ETag options before calling `SetOptionBytes`.

jkralik: Ok. pls create `AddETag`

Danielius1922: Finally created! :D

PR #303, opened 2022-02-01
dtls/tcp/udp: ignore close error during closing connection

jkralik:


tcp/clientconn.go
```diff
@@ -241,7 +240,11 @@ func (cc *ClientConn) Session() *Session {
 
 // Close closes connection without wait of ends Run function.
 func (cc *ClientConn) Close() error {
-	return cc.session.Close()
+	err := cc.session.Close()
+	if coapNet.IsCancelOrCloseError(err) {
```
Danielius1922, line 244: Hmm, something is closing the underlying connection so the atomic variable guard fails sometimes, I guess?

net/error.go
```diff
@@ -1,10 +1,23 @@
...
 
 var ErrListenerIsClosed = io.EOF
 var ErrConnectionIsClosed = io.EOF
 var ErrWriteInterrupted = errors.New("only part data was written to socket")
+
+func IsCancelOrCloseError(err error) bool {
+	if err == nil {
+		return false
+	}
+	if errors.Is(err, context.Canceled) || errors.Is(err, io.EOF) || strings.Contains(err.Error(), "use of closed network connection") {
```
Danielius1922, line 18: `errors.Is(err, net.ErrClosed)` instead of the `strings.Contains` can be used.

tcp/clientconn.go
Danielius1922, line 244: Also I don't think we need to check for Cancel or EOF here, so `errors.Is(err, net.ErrClosed)` should be enough here (and in `udp.Close`).

Danielius1922 (approved):

PR #316, opened 2022-03-29
udp: allows to create ClientConn at server

### LINKED ISSUE #315: How to reuse the udp socket as both Server and Client
Hi @jkralik , I want to make the net.NewListenUDP() and the udp.Dial() uses the same socket so that it can have both server and client functionality.

The reason I want to do this is that I need to forward messages from the local end device to the cloud server, or to forward commands from the cloud server to the end device, for two-way communication.

image

I used net.NewListenUDP() and udp.NewServer() to implement CoAP Server and udp.Dial() to implement CoAP Client.

First, I implement Server(server1, localhost:5683) on the PC to listen to messages from the end device and implement Client (client1) to send these messages to the cloud, for example, 192.168.10.10:12345 post to abcd.com:15683, this step is successful.

Then, when the cloud sends commands to the PC, I need to implement a second Server to listen to the commands from the cloud, such as GET/POST messages. So I listened to this address (server2,192.168.10.10:12345) to receive messages from the cloud, and this step was successful.

The problem I encounter is that when I receive a message from the local end device again, I need to use the same socket as the Client (client1) to send it to the cloud server. But this socket (192.168.10.10:12345) is already listened to and used by server2, so I can't post the message to the cloud server again.

Can you help me with how I can implement listen and dial in the same socket?

Thank you in advance!

jkralik:
Fixes #315

### DIFF
--- a/udp/server.go
+++ b/udp/server.go
@@ -258,16 +258,14 @@ func (s *Server) Serve(l *coapNet.UDPConn) error {
 			case <-s.ctx.Done():
 				return nil
 			default:
+				if errors.Is(err, context.Canceled) || errors.Is(err, io.EOF) || strings.Contains(err.Error(), "use of closed network connection") {
+					return nil
+				}
 				return err
 			}
 		}
 		buf = buf[:n]
-		cc, created := s.getOrCreateClientConn(l, raddr)
-		if created {
-			if s.onNewClientConn != nil {
-				s.onNewClientConn(cc)
-			}
-		}
+		cc := s.getClientConn(l, raddr)
 		err = cc.Process(buf)
 		if err != nil {
 			s.closeConnection(cc)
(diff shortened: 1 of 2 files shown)


IHIDCHAOS (approved):
Can you merge it to the master branch?

udp/server.go
```diff
@@ -258,16 +258,14 @@ func (s *Server) Serve(l *coapNet.UDPConn) error {
 			case <-s.ctx.Done():
 				return nil
 			default:
+				if errors.Is(err, context.Canceled) || errors.Is(err, io.EOF) || strings.Contains(err.Error(), "use of closed network connection") {
```
Danielius1922, line 261: We have function for this -> `net.IsCancelOrCloseError`

Danielius1922 (approved):

PR #327, opened 2022-05-11
message: Implement Options.LocationPath() & Options.SetLocationPath()

maribu:
As trivial as the title says

### DIFF
--- a/message/options.go
+++ b/message/options.go
@@ -33,17 +33,11 @@ func GetPathBufferSize(path string) (int, error) {
 	return size, nil
 }
 
-// SetPath splits path by '/' to URIPath options and copy it to buffer.
-//
-// Return's modified options, number of used buf bytes and error if occurs.
-//
-// @note the url encoded into URIHost, URIPort, URIPath is expected to be
-// absolute (https://www.rfc-editor.org/rfc/rfc7252.txt)
-func (options Options) SetPath(buf []byte, path string) (Options, int, error) {
+func setPath(options Options, optionID OptionID, buf []byte, path string) (Options, int, error) {
 	if len(path) == 0 {
 		return options, 0, nil
 	}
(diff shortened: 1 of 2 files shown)


message/options.go
```diff
@@ -107,10 +107,27 @@ func (options Options) path(buf []byte) (int, error) {
...
+		return "", err
+	}
+	buf = buf[:m]
+	return string(buf), nil
+}
+
+// LocationPath joins Location-Path options by '/' to the buf.
+//
+// Return's number of used buf bytes or error when occurs.
+func (options Options) LocationPath() (string, error) {
```
Danielius1922, line 125: Don't we need a setter or why do only getter is added? 

If I'm reading [rfc7252](https://www.rfc-editor.org/rfc/rfc7252) correctly, LocationPath should be contained in a POST response. If somebody uses go-coap to construct the response we should provide a convenient way to append the option to the message.

jkralik: Yes. Thx

jkralik (approved):

jkralik:
Please, could you add a test? thx

maribu:
Done. Shall I squash?

jkralik:
> Done. Shall I squash?

Yes. Thx

message/options.go
maribu: Sorry for asking again: Yes = squash, or yes = keep them separate?

Danielius1922: Squash it, Jozef wants to squash everything all the time so I'm 100% certain that's what he meant. :D

Danielius1922 (approved):

Danielius1922:
Interesting, we don't have a SonarCloud or codecov reports, I guess the workflows are not setup properly for forked repositories. Note taken.

PR #332, opened 2022-05-24
net/udp: add option WithMulticastInterfaceError:

jkralik:


HRogge:
This is a really nice way to "rescue" the errors from the inner loop.

PR #341, opened 2022-06-18
add TooManyRequests constant to support RFC-8516

tim-golioth:
We recently needed to add this constant locally to our CoAP server, but figured it be good to contribute it upstream as well. I wasn't sure about which uint to assign to it, hence the randomly chosen `145`. Happy to change it to something more appropriate.

### DIFF
--- a/message/codes/code_string.go
+++ b/message/codes/code_string.go
@@ -26,6 +26,7 @@ var codeToString = map[Code]string{
 	PreconditionFailed:    "PreconditionFailed",
 	RequestEntityTooLarge: "RequestEntityTooLarge",
 	UnsupportedMediaType:  "UnsupportedMediaType",
+	TooManyRequests:       "TooManyRequests",
 	InternalServerError:   "InternalServerError",
 	NotImplemented:        "NotImplemented",
 	BadGateway:            "BadGateway",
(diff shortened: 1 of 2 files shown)


beriberikix:
I'm not certain but I think the codes are derived from a decimal conversion. From [RFC7252, Section 5.2](https://datatracker.ietf.org/doc/html/rfc7252#section-5.2):

```
   As a human-readable notation for specifications and protocol
   diagnostics, CoAP code numbers including the Response Code are
   documented in the format "c.dd", where "c" is the class in decimal,
   and "dd" is the detail as a two-digit decimal.  For example,
   "Forbidden" is written as 4.03 -- indicating an 8-bit code value of
   hexadecimal 0x83 (4*0x20+3) or decimal 131 (4*32+3).
```
So would that make 4.29 be 4*32+29 = 157?

tim-golioth:
Thanks for pointing that out, @beriberikix! Totally missed that in the RFC. :facepalm:

jkralik (approved):

beriberikix:
NP! Should we also update the README to include this in the list of RFCs supported?

tim-golioth:
> NP! Should we also update the README to include this in the list of RFCs supported?

Good idea! I'll create a separate PR to update the README.

jkralik:
Thx for contributing 👍

PR #344, opened 2022-06-20
udp: don't return closed connection for server.NewClientConn()

jkralik:


udp/server.go
```diff
@@ -421,11 +421,22 @@ func (s *Server) getOrCreateClientConn(UDPConn *coapNet.UDPConn, raddr *net.UDPA
...
 		if s.onNewClientConn != nil {
 			s.onNewClientConn(cc)
 		}
+		return cc
+	}
+
+	if cc.Context().Err() != nil {
+		// connection is closed so we need to create new one
+		getClose(cc)()
+		return s.getClientConn(l, raddr)
```
Danielius1922, line 439: No chance of infinite recursion, I hope? :-)

udp/server.go
```diff
@@ -421,11 +421,22 @@ func (s *Server) getOrCreateClientConn(UDPConn *coapNet.UDPConn, raddr *net.UDPA
...
 	if created {
 		if s.onNewClientConn != nil {
 			s.onNewClientConn(cc)
 		}
+		return cc
+	}
+
+	if cc.Context().Err() != nil {
+		// connection is closed so we need to create new one
+		getClose(cc)()
```
Danielius1922, line 438: nil check? `getClose` can return nil

udp/server.go
jkralik: fixed 👍

udp/server.go
jkralik: fixed

Danielius1922 (approved):

Danielius1922 (changes requested):
There seems to be some issue with tests right now.

Danielius1922 (approved):

jkralik:
> ( And maybe you can some additional test to increase coverage ;) )

Fixed

PR #352, opened 2022-06-26
Rewrite pkg/cache, pkg/sync/map to use generics

Danielius1922:


pkg/sync/map.go
```diff
@@ -0,0 +1,171 @@
...
+		data: make(map[K]V),
+	}
+}
+
+// Store sets the value for a key.
+func (m *Map[K, V]) Store(key K, value V) {
+	m.StoreWithFunc(key, func() V { return value })
+}
+
+func (m *Map[K, V]) StoreWithFunc(key K, createFunc func() V) {
```
jkralik, line 27: I don't see usability of this function.

pkg/sync/map.go
```diff
@@ -0,0 +1,171 @@
...
+	})
+}
+
+func (m *Map[K, V]) ReplaceWithFunc(key K, onReplaceFunc func(oldValue V, oldLoaded bool) (newValue V, doDelete bool)) (oldValue V, oldLoaded bool) {
+	m.mutex.RLock()
+	v, ok := m.data[key]
+	m.mutex.RUnlock()
+	newValue, del := onReplaceFunc(v, ok)
+	m.mutex.Lock()
+	defer m.mutex.Unlock()
```
jkralik, line 116: this need to be guard with `m.mutex.Lock()` otherwise some goroutine can use replaced object at same time.

pkg/sync/map.go
```diff
@@ -0,0 +1,171 @@
...
+		if onLoadFunc != nil {
+			v = onLoadFunc(v)
+		}
+		return v, ok
+	}
+	v = createFunc()
+	m.mutex.Lock()
+	m.data[key] = v
+	m.mutex.Unlock()
+	return v, ok
```
jkralik, line 70: this need to be guard with `m.mutex.Lock()` otherwise some goroutine can use create object at same time.

pkg/sync/map.go
Danielius1922: I've copied the interface from https://github.com/plgd-dev/kit/blob/main/sync/map.go, but yeah in general the 'withFunc' methods don't seem all that useful.

pkg/sync/map.go
Danielius1922: What do you mean - even if I lock this whole method the lock will be unlocked when method ends. So multiple goroutines will still access the object concurrently, but that should be handled by the object itself.

jkralik (approved):

PR #357, opened 2022-07-11
merge {tcp,udp}/message to message

jkralik:


dtls/client_test.go
```diff
@@ -163,6 +163,7 @@ func TestClientConnGetSeparateMessage(t *testing.T) {
 
 	m := mux.NewRouter()
 	err = m.Handle("/a", mux.HandlerFunc(func(w mux.ResponseWriter, r *mux.Message) {
+		r.Hijack()
```
Danielius1922, line 166: Doesn't hijacked message need to be manually released in the goroutine?

tcp/clientconn.go
```diff
@@ -132,11 +132,16 @@ func bwCreateReleaseMessage(messagePool *pool.Pool) func(m blockwise.Message) {
 	}
 }
 
+type observationMessage struct {
+	ctx context.Context
+	message.Message
+}
+
 func bwCreateHandlerFunc(messagePool *pool.Pool, observationRequests *kitSync.Map) func(token message.Token) (blockwise.Message, bool) {
```
Danielius1922, line 140: Is this on the latest master? I think I've removed almost all usage of `kitSync` pkg.

dtls/client_test.go
jkralik: No it is not required, but allocated message is not returned back to pool. (for performance reason it will be nice if we return it back.) but in test it is not needed.

tcp/clientconn.go
jkralik: Fixed

message/pool/message.go
```diff
@@ -387,3 +468,54 @@ func (r *Message) ReadBody() ([]byte, error) {
...
+		return nil, err
+	}
+	if len(r.bufferMarshal) < size {
+		r.bufferMarshal = append(r.bufferMarshal, make([]byte, size-len(r.bufferMarshal))...)
+	}
+	n, err := encoder.Encode(msg, r.bufferMarshal)
+	if err != nil {
+		return nil, err
+	}
+	r.bufferMarshal = r.bufferMarshal[:n]
```
Danielius1922, line 498: Is this an optimization? Why not always allocate the returned array? 

Couldn't something like this happen:
```golang
data1 := r.MarshalWithEncoder(encoder1)
...
data2 := r.MarshalWithEncoder(encoder2)

...do something with data1...
```

If the buffer from the first call is large enough for the second call, then data1 will be corrupted.

message/pool/message.go
```diff
@@ -1,45 +1,126 @@
...
+	if len(message.Payload) > 0 {
+		r.body = bytes.NewReader(message.Payload)
+	}
+	r.isModified = true
+}
+
+// UpsertMessageID set value only when origin value is invalid. Only 0 to 2^16-1 values are valid.
+func (r *Message) UpsertMessageID(mid int32) {
+	if r.msg.MessageID >= 0 && r.msg.MessageID < math.MaxUint16 {
+		return
```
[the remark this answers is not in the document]
jkralik: No, this is used as fallback when it is not set by developer.

message/pool/pool.go
```diff
@@ -0,0 +1,55 @@
...
+// The returned Message instance may be passed to ReleaseMessage when it is
+// no longer needed. This allows Message recycling, reduces GC pressure
+// and usually improves performance.
+func (p *Pool) AcquireMessage(ctx context.Context) *Message {
+	v := p.messagePool.Get()
+	if v == nil {
+		return NewMessage(ctx)
+	}
+	r := v.(*Message)
+	p.currentMessagesInPool.Dec()
```
Danielius1922, line 37: Wouldn't this become negative when you invoke `AcquireMessage` twice before invoking `ReleaseMessage`? In that case the counter wouldn't work.

Example:
```
p := pool.New(1, 1024)

p.AcquireMessage(ctx) // p.currentMessagesInPool == -1
p.AcquireMessage(ctx) // p.currentMessagesInPool == -2
p.ReleaseMessage(req) // p.currentMessagesInPool == -1, req is put into the pool
p.ReleaseMessage(req) // p.currentMessagesInPool == 0, req is put into the pool, so more messages are in the pool than allowed
```

message/pool/pool.go
```diff
@@ -0,0 +1,55 @@
...
+	r.ctx = ctx
+	return r
+}
+
+// ReleaseMessage returns req acquired via AcquireMessage to Message pool.
+//
+// It is forbidden accessing req and/or its' members after returning
+// it to Message pool.
+func (p *Pool) ReleaseMessage(req *Message) {
+	v := p.currentMessagesInPool.Load()
```
Danielius1922, line 47: There's a race condition here if two threads use the same pool (say with a `pool` with `maxNumMessages` = 1):

Thread 1:
```
v := p.currentMessagesInPool.Load() // v == 0
// thread is stopped here and Thread 2 is scheduled
```

Thread 2:
```
v := p.currentMessagesInPool.Load() // v == 0
...
req is put into the message pool
```
Thread 2 now resumes and we have 2 messages in the pool. 

Good ol' compare and swap loop must be used here for correctness:
```go
for true {
  v := p.currentMessagesInPool.Load()
  if v >= int64(p.maxNumMessages) {
    return
  }
  next := v + 1
  if (p.currentMessagesInPool.CAS(v, next)) {
    break
  }
}
```

message/pool/pool.go
jkralik: No. `p.currentMessagesInPool` is incremented when message is released.
```
p := pool.New(1, 1024)
p.AcquireMessage(ctx) // p.currentMessagesInPool = 0
p.AcquireMessage(ctx) // p.currentMessagesInPool = 0
p.AcquireMessage(ctx) // p.currentMessagesInPool = 0
p.ReleaseMessage(req) // p.currentMessagesInPool = 1
```

Danielius1922: `p.messagePool.Get()` will return nil, because no message has been put into the pool yet, right? Indeed, this should be correct then.

message/pool/pool.go
jkralik: fixed

udp/client/clientconn.go
```diff
@@ -297,8 +297,10 @@ func (cc *ClientConn) WriteMessage(req *pool.Message) error {
 		req := bwreq.(*pool.Message)
 		if req.Options().HasOption(message.Block1) || req.Options().HasOption(message.Block2) {
 			req.SetMessageID(cc.getMID())
```
Danielius1922, line 299: Shouldn't now the return value of getMID be checked, since it now returns int32? This could create a message with invalid ID if somebody uses a bad ID generator.

jkralik: This is a user responsibility, but if user set invalid message then coder returns error during encoding:
https://github.com/plgd-dev/go-coap/blob/4e633a457ee7c1772e8b78c191a7b3c594ae9755/udp/coder/coder.go#L50

PR #360, opened 2022-07-19
merge observation handler

jkralik:


udp/client/clientconn.go
```diff
@@ -614,8 +643,8 @@ func (cc *ClientConn) getResponseFromCache(mid int32, resp *pool.Message) (bool,
 // CheckMyMessageID compare client msgID against peer messageID and if it is near < 0xffff/4 then incrase msgID.
 // When msgIDs met it can cause issue because cache can send message to which doesn't bellows to request.
 func (cc *ClientConn) CheckMyMessageID(req *pool.Message) {
-	if req.Type() == message.Confirmable && uint16(req.MessageID())-uint16(atomic.LoadUint32(&cc.msgID)) < 0xffff/4 {
-		atomic.AddUint32(&cc.msgID, 0xffff/2)
+	if req.Type() == message.Confirmable && uint16(req.MessageID())-uint16(cc.msgID.Load()) < 0xffff/4 {
+		cc.msgID.Add(0xffff / 2)
```
Danielius1922, line 647: Time of check vs time of use race:

ThreadA executes line 646 and get rescheduled, ThreadB does the same and both increase the msgID by `0xffff / 2`. Don't know if that's a problem, if it is we need to use CAS here:

```go
if req.Type() == message.Confirmable {
  for {  
    oldID := cc.msgID.Load()
    if uint16(req.MessageID())-uint16(cc.msgID.Load()) >= 0xffff/4 {
      return
    }
    newID := oldID + 0xffff / 2
    if cc.msgID.CAS(oldID, newID) {
      break
    }
  }

}

```


Also msgID can be -1 or some int32_t value that's invalid for int16_t, right? Shouldn't that be checked for?

net/observation/handler.go
```diff
@@ -0,0 +1,230 @@
...
+// Observation represents subscription to resource on the server
+type Observation[C Client] struct {
+	req                 message.Message
+	observeFunc         func(req *pool.Message)
+	respObservationChan chan respObservationMessage
+	waitForResponse     atomic.Bool
+	observationHandler  *Handler[C]
+
+	mutex       sync.Mutex
+	obsSequence uint32    // guarded by mutex
```
Danielius1922, line 133: I liked the `private struct` convention used in some other places.

net/observation/handler.go
```diff
@@ -0,0 +1,230 @@
...
+		waitForResponse:     *atomic.NewBool(true),
+		respObservationChan: respObservationChan,
+		observeFunc:         observeFunc,
+		observationHandler:  observationHandler,
+	}
+}
+
+func (o *Observation[C]) handle(w *responsewriter.ResponseWriter[C], r *pool.Message) {
+	code := r.Code()
+	notSupported := !r.HasOption(message.Observe)
```
Danielius1922, line 155: lines 154-155 can be moved after the if

udp/client/clientconn.go
jkralik: `msgID` is uint32 ab it is always cast to uint16

Danielius1922 (approved):

PR #361, opened 2022-07-19
allow to get response message from ResponseWriter

jkralik:


jkralik:
v3 contains the functions that allow you to modify message directly:
https://github.com/plgd-dev/go-coap/blob/b8901963ae43acd9a14f98b3d47ec17b26b38e7a/net/responsewriter/responseWriter.go#L59
https://github.com/plgd-dev/go-coap/blob/b8901963ae43acd9a14f98b3d47ec17b26b38e7a/net/responsewriter/responseWriter.go#L65

PR #363, opened 2022-07-20
use generics in BW transfer

jkralik:


net/blockwise/blockwise.go
```diff
@@ -193,20 +159,18 @@ func newRequestGuard(request Message) *messageGuard {
 
 // NewBlockWise provides blockwise.
 // getSentRequestFromOutside must returns a copy of request which will be released by function releaseMessage after use.
```
Danielius1922, line 161: NewBlockWise -> New and is that bit about releaseMessage still valid?

jkralik: fixed

PR #365, opened 2022-07-21
merge client/server options

jkralik:


tcp/client/clientconn.go
```diff
@@ -18,226 +17,69 @@ import (
...
-}
-
 // NewClientConn creates connection over session and observation.
-func NewClientConn(ctx context.Context,
+func NewClientConn(
 	connection *coapNet.Conn,
+	createBlockWise func(cc *ClientConn) *blockwise.BlockWise[*ClientConn],
+	inactivityMonitor inactivity.Monitor,
+	cfg *Config,
+	/*ctx context.Context,
```
Danielius1922, line 46: rm?

options/commonOptions.go
```diff
@@ -0,0 +1,593 @@
...
+	}
+}
+
+func (o HandlerFuncOpt[H]) UDPClientApply(cfg *udpClient.Config) {
+	switch v := any(o.h).(type) {
+	case udpClient.HandlerFunc:
+		cfg.Handler = v
+	default:
+		var t udpClient.HandlerFunc
+		panic(fmt.Errorf("invalid HandlerFunc type %T, expected %T", v, t))
```
Danielius1922, line 81: panicForInvalidHandlerFunc ?

options/commonOptions.go
```diff
@@ -0,0 +1,593 @@
...
+}
+
+func panicForInvalidHandlerFunc(t, exp any) {
+	panic(fmt.Errorf("invalid HandlerFunc type %T, expected %T", t, exp))
+}
+
+func (o HandlerFuncOpt[H]) TCPServerApply(cfg *tcpServer.Config) {
+	switch v := any(o.h).(type) {
+	case tcpClient.HandlerFunc:
+		cfg.Handler = v
```
Danielius1922, line 38: We could create a base config struct with all the shared members by `tcpServer.Config`, `tcpClient.Config`, `udpServer.Config`, `dtlsServer.Config` and `udpClient.Config`. It seems to be something like this:

```go
type CommonConfig struct {
	Ctx                            context.Context
	Handler                        HandlerFunc
	Errors                         ErrorFunc
	GoPool                         GoPoolFunc
	PeriodicRunner                 periodic.Func
	MessagePool                    *pool.Pool
	GetToken                       client.GetTokenFunc
	CreateInactivityMonitor        func() inactivity.Monitor
	MaxMessageSize                 uint32
	BlockwiseTransferTimeout       time.Duration
	BlockwiseSZX                   blockwise.SZX	
	BlockwiseEnable                bool
}
```

options/commonOptions.go
jkralik: fixed

tcp/client/clientconn.go
jkralik: removed

options/commonOptions.go
jkralik: Fixed

udp/client_test.go
```diff
@@ -124,7 +124,7 @@ func TestClientConnGet(t *testing.T) {
 	require.NoError(t, err)
 	err = m.Handle("/empty", mux.HandlerFunc(func(w mux.ResponseWriter, r *mux.Message) {
 		assert.Equal(t, codes.GET, r.Code())
-		// Calling SetResponse was failing with an EOF error when the reader is empty
+		// Calling SetResponse was failing options.With an EOF error when the reader is empty
```
Danielius1922, line 127: revert

options/commonOptions.go
```diff
@@ -0,0 +1,666 @@
...
+func WithHandlerFunc[H Handler](h H) HandlerFuncOpt[H] {
+	return HandlerFuncOpt[H]{h: h}
+}
+
+// HandlerFuncOpt handler function option.
+type MuxHandlerOpt struct {
+	m mux.Handler
+}
+
+func (o MuxHandlerOpt) TCPServerApply(cfg *tcpServer.Config) {
```
Danielius1922, line 100: When they implement generic methods we can get rid of these overload and keep a single method.

tcp/server/config.go
```diff
@@ -0,0 +1,67 @@
...
+
+type GoPoolFunc = func(func()) error
+
+// OnNewClientConnFunc is the callback for new connections.
+type OnNewClientConnFunc = func(cc *client.ClientConn)
+
+var DefaultConfig = func() Config {
+	opts := Config{
+		Common: config.DefaultCommon(),
+		CreateInactivityMonitor: func() client.InactivityMonitor {
```
Danielius1922, line 35: Is this deliberate? Previously, the default was a nil monitor.

tcp/server.go:
```go
createInactivityMonitor: func() inactivity.Monitor {
    return inactivity.NewNilMonitor()
},
```

tcp/server/server.go
```diff
@@ -0,0 +1,229 @@
...
+			// default no-op
+		}
+	}
+	if cfg.GetToken == nil {
+		cfg.GetToken = message.GetToken
+	}
+	errorsFunc := cfg.Errors
+	// assign updated func to opts.errors so opts.handler also uses the updated error handler
+	cfg.Errors = func(err error) {
+		if errors.Is(err, context.Canceled) || errors.Is(err, io.EOF) || strings.Contains(err.Error(), "use of closed network connection") {
```
Danielius1922, line 69: we have `net.IsCancelOrCloseError` for this check, should be substituted for all remaining code with 

```go
strings.Contains(err.Error(), "use of closed network connection"
```

tcp/server/config.go
jkralik: Yes, it is intentional. TCP Server want's to watch if clients are reachable by default.

tcp/server/server.go
jkralik: fixed

Danielius1922 (approved):

PR #366, opened 2022-07-22
Fix for messages coming from the message pool always ends up as non-confirmable

andersroos:
Hello!

While using this lib I discovered that observe requests sometimes wasn't retried properly ([here](https://github.com/plgd-dev/go-coap/blob/e59fa916285a183e0adbcbf6d0c2e1329ddc448e/udp/client/clientconn.go#L258)). After a bit of digging I found out that messages coming from the message pool always are set to non-confirmable while new messages not coming from the pool are confirmable by default.

Since the blockwise code copies messages it can end up sending a non-confirmable message when the original message was actually set as confirmable.

I am not very familiar with this code base so I am not sure this fix is good, I have tested it with our server and it seems to solve the issue.

To me it feels wrong to assume what the type should be, so I guess it needs to be set every time a message is acquired? Regardless if it is reused or new? I haven't checked the reset of the code base to see if there are more places like this, just this file. In the meantime I use a message pool with 0 max size as a workaround since most messages should be conformable anyway, but that could cause other issues I suppose, I mean I assume that reset sets a message to non-confirmable for a reason.

Cheers and thanks for a good library!

### DIFF
--- a/net/blockwise/blockwise.go
+++ b/net/blockwise/blockwise.go
@@ -491,6 +491,10 @@ func (b *BlockWise) sendEntityIncomplete(w ResponseWriter, r Message, token mess
 	sendMessage := b.acquireMessage(w.Message().Context())
 	sendMessage.SetCode(codes.RequestEntityIncomplete)
 	sendMessage.SetToken(token)
+	if msg, ok := sendMessage.(hasType); ok {
+		msg.SetType(udpMessage.NonConfirmable)
+	}
+
 	setTypeFrom(sendMessage, r)
 	w.SetMessage(sendMessage)
 }
@@ -900,7 +904,6 @@ func (b *BlockWise) processReceivedMessage(w ResponseWriter, r Message, maxSzx S
 	szx = getSzx(szx, maxSzx)
 	sendMessage := b.acquireMessage(r.Context())
 	sendMessage.SetToken(token)
-	// TODO What to set type to?
 	if blockType == message.Block2 {
 		num = payloadSize / szx.Size()
 		sendMessage.ResetOptionsTo(sentRequest.Options())
--- a/net/blockwise/blockwise.go
+++ b/net/blockwise/blockwise.go
@@ -494,8 +494,6 @@ func (b *BlockWise) sendEntityIncomplete(w ResponseWriter, r Message, token mess
 	if msg, ok := sendMessage.(hasType); ok {
 		msg.SetType(udpMessage.NonConfirmable)
 	}
-
-	setTypeFrom(sendMessage, r)
 	w.SetMessage(sendMessage)
 }
 
--- a/net/blockwise/blockwise.go
+++ b/net/blockwise/blockwise.go
@@ -487,7 +487,7 @@ func (b *BlockWise) RemoveFromResponseCache(token message.Token) {
 	b.sendingMessagesCache.Delete(token.Hash())
 }
 
-func (b *BlockWise) sendEntityIncomplete(w ResponseWriter, r Message, token message.Token) {
+func (b *BlockWise) sendEntityIncomplete(w ResponseWriter, token message.Token) {
 	sendMessage := b.acquireMessage(w.Message().Context())
 	sendMessage.SetCode(codes.RequestEntityIncomplete)
 	sendMessage.SetToken(token)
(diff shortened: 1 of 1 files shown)


Danielius1922:
(Seems OK to me, but I'll wait for Jozef to go over it. He's enjoying his holiday this week, so that'll have to wait for next week.)

net/blockwise/blockwise.go
```diff
@@ -889,6 +894,7 @@ func (b *BlockWise) processReceivedMessage(w ResponseWriter, r Message, maxSzx S
 	szx = getSzx(szx, maxSzx)
 	sendMessage := b.acquireMessage(r.Context())
 	sendMessage.SetToken(token)
+	// TODO What to set type to?
```
jkralik, line 897: In `v2` this is handled by:
https://github.com/plgd-dev/go-coap/blob/e59fa916285a183e0adbcbf6d0c2e1329ddc448e/udp/client/clientconn.go#L220
and
https://github.com/plgd-dev/go-coap/blob/e59fa916285a183e0adbcbf6d0c2e1329ddc448e/udp/client/clientconn.go#L296

for `v3` this will be refactored to block-wise.

net/blockwise/blockwise.go
```diff
@@ -482,10 +484,11 @@ func (b *BlockWise) RemoveFromResponseCache(token message.Token) {
 	b.sendingMessagesCache.Delete(token.Hash())
 }
 
-func (b *BlockWise) sendEntityIncomplete(w ResponseWriter, token message.Token) {
+func (b *BlockWise) sendEntityIncomplete(w ResponseWriter, r Message, token message.Token) {
 	sendMessage := b.acquireMessage(w.Message().Context())
 	sendMessage.SetCode(codes.RequestEntityIncomplete)
 	sendMessage.SetToken(token)
+	setTypeFrom(sendMessage, r)
```
jkralik, line 491: This is not OK. You cannot send  response with `Confirmable` for `Confirmable` request.  I don't think you need to  add this line.

andersroos:
Thanks @jkralik! Is on a vacation trip at the moment, will take this up again when I am back home.

net/blockwise/blockwise.go
andersroos: Hmm, I don't really get this, how is it handled by those lines, can you help me understand? I mean eventually the execution reaches this code

https://github.com/plgd-dev/go-coap/blob/948a0bcf18a66f92d7269f40104a64e997498d3b/udp/client/clientconn.go#L233 

and the message type is used to determine if there should be retries or not. So unless the type is set explicitly somewhere it will be random, `Confirmable` if it is a new message created in the message pool or it will be `NonConfirmable` if it is a reused message from the message pool.

net/blockwise/blockwise.go
andersroos: Ok, maybe it is safer to always set it to `NonConfirmable` then? According to the spec `Confirmable` responses are allowed, but maybe they are unsuitable in this implementation? `NonConfirmable` responses are also always allowed if I read the spec correctly.

https://www.rfc-editor.org/rfc/rfc7252#section-5.2.3

jkralik: Yeh, this is fine. I'm preparing the new go-coap/v3 in https://github.com/plgd-dev/go-coap/issues/368 for the October release. Please is it possible to migrate this fix to branch [v3](https://github.com/plgd-dev/go-coap/tree/v3)?

andersroos: @jkralik Changed to always `NonConfirmable`.

I haven't looked at the v3 branch yet, but I can take a stab at migrating it.

net/blockwise/blockwise.go
```diff
@@ -482,10 +484,15 @@ func (b *BlockWise) RemoveFromResponseCache(token message.Token) {
...
-func (b *BlockWise) sendEntityIncomplete(w ResponseWriter, token message.Token) {
+func (b *BlockWise) sendEntityIncomplete(w ResponseWriter, r Message, token message.Token) {
 	sendMessage := b.acquireMessage(w.Message().Context())
 	sendMessage.SetCode(codes.RequestEntityIncomplete)
 	sendMessage.SetToken(token)
+	if msg, ok := sendMessage.(hasType); ok {
+		msg.SetType(udpMessage.NonConfirmable)
+	}
+
+	setTypeFrom(sendMessage, r)
```
jkralik, line 495: Is still needed this line ?

andersroos: @jkralik No, sorry, fixed now.

jkralik:
@andersroos Pls could you resolve linter issue - https://github.com/plgd-dev/go-coap/actions/runs/3014872294 ? After that I will merge PR.

andersroos:
> @andersroos Pls could you resolve linter issue - https://github.com/plgd-dev/go-coap/actions/runs/3014872294 ? After that I will merge PR.

Sure! Done.

jkralik (approved):

jkralik:
Great. Thx for contributing !

PR #376, opened 2022-08-04
nstart - max number of simultaneous outstanding requests to the server

jkralik:


udp/client/clientconn.go
```diff
@@ -246,6 +285,11 @@ func (cc *ClientConn) writeMessage(req *pool.Message) error {
 		defer func() {
 			_, _ = cc.midHandlerContainer.PullOut(req.MessageID())
 		}()
+	case message.NonConfirmable:
+		/* TODO need to acquireOutstandingInteraction
```
Danielius1922, line 289: TODO for the future?

jkralik: Yes

udp/client/clientconn.go
```diff
@@ -226,13 +238,40 @@ func (cc *ClientConn) Do(req *pool.Message) (*pool.Message, error) {
...
+func (cc *ClientConn) releaseOutstandingInteraction() {
+	cc.numOutstandingInteraction.Release(1)
+}
+
+func (cc *ClientConn) acquireOutstandingInteraction(ctx context.Context) error {
+	nStart := cc.Transmission().nStart.Load()
+	if nStart == 0 {
+		return fmt.Errorf("invalid NStart value %v", nStart)
+	}
+	n := math.MaxInt64 - int64(cc.Transmission().nStart.Load()) + 1
```
Danielius1922, line 250: Do we have a documentation for the nStart property? We should add a note somewhere that changing the nStart value will only effect requests queued after the change (and the requests waiting here already before the change will get unblocked when enough weight has been released based on the nStart value when they were queued and won't get unblocked because the value increased).

Danielius1922 (approved):

udp/client/clientconn.go
jkralik: Ok so I changed the behavior -> `acquireOutstandingInteraction` will use loop to get access.

jkralik:
Golangci-lint - is fixed in another PR

Danielius1922 (approved):

udp/client/clientconn_test.go
```diff
@@ -280,28 +282,45 @@ func TestClientConnGet(t *testing.T) {
...
 		})
 	}
 }
 
+func TestParallelClientConnGet(t *testing.T) {
+	testParallelClientConnGet(t, testNumParallel)
+}
+
+/*
+	func TestClientConnGet(t *testing.T) {
```
Danielius1922, line 320: rm?

udp/client/clientconn_test.go
```diff
@@ -482,29 +501,47 @@ func TestClientConnPost(t *testing.T) {
...
 }
 
-func TestClientConnPut(t *testing.T) {
+func TestClientConnPost(t *testing.T) {
+	testClientConnPost(t, 1)
+}
+
+/*
+func TestParallelClientConnPost(t *testing.T) {
+	testClientConnPost(t, testNumParallel)
```
Danielius1922, line 540: Are the commented out tests not working or what is the problem?

udp/client/clientconn_test.go
jkralik: no - uncommented :)

udp/client/clientconn_test.go
jkralik: Parallel request were sync with blockwise, but in blockwise has been removed so this test fails. I think we need to solve it by another way.

PR #378, opened 2022-08-05
allow to observe with customized request

jkralik:


tcp/server/config.go
```diff
@@ -13,8 +13,8 @@ import (
 	"github.com/plgd-dev/go-coap/v3/tcp/client"
 )
 
-// A ServerOption sets options such as credentials, codec and keepalive parameters, etc.
-type TCPServerOption interface {
+// A Option sets options such as credentials, codec and keepalive parameters, etc.
+type TCPOption interface {
```
Danielius1922, line 17: What's the difference between `TCPOption` and `Option` from `tcp/server/server.go` ? Both of these have the same single method - `TCPServerApply(*Config)`. Seems like duplicity.

jkralik: This was duplicity - removed

Danielius1922 (approved):

PR #381, opened 2022-08-12
fix leaking of pool.Message

jkralik:


net/blockwise/blockwise_test.go
```diff
@@ -303,15 +304,20 @@ func TestBlockWiseDo(t *testing.T) {
...
+			require.Len(t, receivingMessagesCache, 0)
+			sendingMessagesCache := sender.sendingMessagesCache.PullOutAll()
+			require.Len(t, sendingMessagesCache, 0)
 			require.NoError(t, err)
 			assert.Equal(t, tt.want, fromPoolMessage(got))
 		})
 	}
 }
 
+/*
```
Danielius1922, line 317: The parallel test doesn't work anymore?

jkralik: No this need to be fixed separately.

net/blockwise/blockwise.go
```diff
@@ -429,6 +380,23 @@ func (b *BlockWise[C]) sendEntityIncomplete(w *responsewriter.ResponseWriter[C],
 	w.SetMessage(sendMessage)
 }
 
+func wantToReceived(r *pool.Message) bool {
```
Danielius1922, line 383: [nitpick - bad grammar] `wantsToReceive` or `wantsToBeReceived`

net/blockwise/blockwise.go
```diff
@@ -462,8 +430,8 @@ func (b *BlockWise[C]) Handle(w *responsewriter.ResponseWriter[C], r *pool.Messa
 		b.errors(fmt.Errorf("continueSendingMessage(%v): %w", r, err))
 		return
 	}
-	if b.autoCleanUpResponseCache && !more {
-		b.RemoveFromResponseCache(token)
+	if !more && sendingMessageCached.Data().Code() > codes.DELETE {
```
Danielius1922, line 433: Are codes `>` than `codes.DELETE` invalid or why are only those codes deleted from cache?

Danielius1922 (approved):

net/blockwise/blockwise.go
jkralik: For codes GET,POST,PUT,DELETE, we want them to wait for pairing response and then delete them when the full response comes in or when timeout occurs.

net/blockwise/blockwise.go
jkralik: fixed

Issue #384, opened 2022-08-22
Comment on SetObserve method is confusing


tim-golioth:
I can submit a PR for this, but I wanted to check first that I'm not reading the code incorrectly.

The comment for the [SetObserve](https://github.com/plgd-dev/go-coap/blob/948a0bcf18a66f92d7269f40104a64e997498d3b/message/options.go#L338) method reads:

```
// SetObserve sets ContentFormat option.
func (options Options) SetObserve(buf []byte, observe uint32) (Options, int, error) {
...
```

That should be `SetObserve sets the Observe option.` Correct?

Danielius1922:
Yes, must have been a copy & paste error on my part.

PR #390, opened 2022-09-02
Upgrade dependencies

Danielius1922:
Direct:
github.com/stretchr/testify v1.8.0
go.uber.org/atomic v1.10.0
golang.org/x/exp v0.0.0-20220827204233-334a2380cb91
golang.org/x/net v0.0.0-20220826154423-83b083e8dc8b
golang.org/x/sync v0.0.0-20220819030929-7fc1605a5dde

Indirect:
gopkg.in/yaml.v3 v3.0.1

PR #391, opened 2022-09-05
fix comment on function SetObserve

AryanGodara:
Signed-off-by: aryan <aryangodara03@gmail.com>
With reference to issue #384

### DIFF
--- a/message/options.go
+++ b/message/options.go
@@ -335,7 +335,7 @@ func (options Options) SetContentFormat(buf []byte, contentFormat MediaType) (Op
 	return options.SetUint32(buf, ContentFormat, uint32(contentFormat))
 }
 
-// SetObserve sets ContentFormat option.
+// SetObserve sets Observe option.
 func (options Options) SetObserve(buf []byte, observe uint32) (Options, int, error) {
 	return options.SetUint32(buf, Observe, observe)
 }


Danielius1922 (approved):

AryanGodara:
@Danielius1922 , Should I fix the issues, to pass the `lint` test?
I'm asking before implementing, since it can complicate the PR, by making those other chanes to the code, if it's not necessary.
But if needed, I'll be glad to do it.

Danielius1922:
> @Danielius1922 , Should I fix the issues, to pass the `lint` test? I'm asking before implementing, since it can complicate the PR, by making those other chanes to the code, if it's not necessary. But if needed, I'll be glad to do it.

Hey, thanks for the fix, I think I've fixed those and other issues in PR #390, so it's already done. Thus nothing here needs to be done, I'll merge.

AryanGodara:
@Danielius1922 , I had a request.
I use this library `go-coap` in my code for coap-adapter. And I want to create unit tests for coap-handler function.
So, does this library offer mocking features, kind of like the `httptest.NewServer()` for http.
I was thinking of writing a mock-coap server, but that is way out of my league for now, and I kind of only have 2 weeks to finish the task
Or maybe you could suggest a better way to write tests for coap-handlers using `go-coap`.

Danielius1922:
> @Danielius1922 , I had a request. I use this library `go-coap` in my code for coap-adapter. And I want to create unit tests for coap-handler function. So, does this library offer mocking features, kind of like the `httptest.NewServer()` for http. I was thinking of writing a mock-coap server, but that is way out of my league for now, and I kind of only have 2 weeks to finish the task Or maybe you could suggest a better way to write tests for coap-handlers using `go-coap`.

No mocking features like those offered by httptest are available currently. I've looked at the source of the httptest library and it's powerful yet small (only 3 implementation files), so creating something like that is possible. We've talked about this with @jkralik and we both agree it would be a nice enhancement, but it's unlikely we will get to it any time soon, the plan is to finish v3 first and then we can look into new features. 

So currently the only way I can think of doing what you need is: create a newServer function that makes a coap server running on localhost on some port in a separate go-routine with selected uri handlers. Then call newServer at the start of each client test case. But this is basically the implementation of httptest equivalent. If you go that way you might as well contribute it here, I might have some time to help (usually I have few hours on a Friday to spare).

(Not much of a help I know, sorry. BTW nice avatar, I hope HxH doesn't end up unfinished like Berserk.)

PR #396, opened 2022-09-21
allow to limit client parallel requests

jkralik:


net/client/limitParallelRequests/limitParallelRequests.go
```diff
@@ -0,0 +1,135 @@
...
+
+	"github.com/plgd-dev/go-coap/v3/message"
+	"github.com/plgd-dev/go-coap/v3/message/pool"
+	coapSync "github.com/plgd-dev/go-coap/v3/pkg/sync"
+	"golang.org/x/sync/semaphore"
+)
+
+type (
+	DoFunc        = func(req *pool.Message) (*pool.Message, error)
+	DoObserveFunc = func(req *pool.Message, observeFunc func(req *pool.Message), opts ...message.Option) (Observation, error)
```
Danielius1922, line 17: DoFunc and DoObserveFunc are already defined in [net/client/client.go], seems duplicit.

Danielius1922 (approved):

net/client/limitParallelRequests/limitParallelRequests.go
jkralik: Removed from `net/client/client.go`

PR #398, opened 2022-09-26
merge master to v3

jkralik:


.golangci.yml
```diff
@@ -117,5 +117,15 @@ issues:
       linters:
         - gosec
 
+    - path: example_test\.go
+      text: "exitAfterDefer"
+      linters:
+        - gocritic
+
+    - path: udp/message/getmid.go
```
Danielius1922, line 125: I think this path is not relevant for v3 branch, so this whole item can be removed.

Danielius1922 (approved):

PR #402, opened 2022-10-03
Add expiration time customization to the cache

lubo-svk:


pkg/cache/cache.go
```diff
@@ -4,34 +4,36 @@ import (
...
-	return now.After(e.validUntil)
+	return now.After(value)
 }
 
 func (e *Element) Data() interface{} {
 	return e.data
 }
 
-func NewElement(data interface{}, validUntil time.Time, onExpire func(d interface{})) *Element {
+func NewElement(data interface{}, validUntil *atomic.Time, onExpire func(d interface{})) *Element {
```
jkralik, line 32: this doesn't need to be `*atomic.Time` just `time.Time`

pkg/cache/cache.go
```diff
@@ -4,34 +4,36 @@ import (
...
+	"go.uber.org/atomic"
 )
 
 func DefaultOnExpire(d interface{}) {
 	// for nothing on expire
 }
 
 type Element struct {
-	validUntil time.Time
+	ValidUntil *atomic.Time
```
jkralik, line 15: it could be without pointer `atomic.Time`

jkralik (approved):

PR #405, opened 2022-10-07
release v3

### LINKED ISSUE #368: preparing new v3 release
## v3 main branch

https://github.com/plgd-dev/go-coap/tree/v3

## Features:

- [x] use only one pool.Message for tcp,udp,dtls - #325
- [x] allow get MessageID and Type from pool.Message
- [x] allow set MessageID and Type to pool.Message
- [x] https://github.com/plgd-dev/go-coap/issues/140
- [x] https://github.com/plgd-dev/go-coap/issues/312
- [x] https://github.com/plgd-dev/go-coap/issues/251
- [x] https://github.com/plgd-dev/go-coap/issues/226
- [x] update go-coap version to v3
- [x] https://github.com/plgd-dev/go-coap/issues/383
- [x] remove duplicate code as match as possible
- [x] ...

## Optional features:
- [ ] https://github.com/plgd-dev/go-coap/issues/313

## Release date
2022.10

jkralik:
Fixes #368


go.mod
```diff
@@ -1,12 +1,12 @@
...
 go 1.18
 
 require (
 	github.com/dsnet/golib/memfile v1.0.0
-	github.com/pion/dtls/v2 v2.1.5
-	github.com/plgd-dev/kit/v2 v2.0.0-20211006190727-057b33161b90
-	github.com/stretchr/testify v1.8.0
+	github.com/pion/dtls/v2 v2.1.6-0.20220815090156-a04cfccb23e5
+	github.com/pion/udp v0.1.1
+	github.com/stretchr/testify v1.7.1
```
Danielius1922, line 9: This seems weird to me, why is the version here (v1.7.1) older than the previous one (v1.8.0)?

jkralik: fixed

jkralik:
Blocked for PR: https://github.com/pion/udp/pull/75

tcp/client/session.go
```diff
@@ -165,123 +149,96 @@ func (s *Session) Close() error {
 }
 
 func (s *Session) Sequence() uint64 {
-	return atomic.AddUint64(&s.sequence, 1)
+	return s.sequence.Inc()
 }
 
 func (s *Session) Context() context.Context {
-	return *s.ctx.Load().(*context.Context)
+	return *s.ctx.Load().(*context.Context) //nolint:forcetypeassert
```
Danielius1922, line 156: This seems like a data race, in `SetContextValue` above `s.ctx` is protected by the `s.mutex`. But the read here is not protected.

net/dtlslistener.go
```diff
@@ -39,14 +92,56 @@ func NewDTLSListener(network string, addr string, dtlsCfg *dtls.Config) (*DTLSLi
...
+}
+
+func (l *DTLSListener) run() {
+	defer l.wg.Done()
+	for {
+		if l.closed.Load() {
+			return
+		}
+		err := l.accept()
+		if errors.Is(err, udp.ErrClosedListener) {
```
Danielius1922, line 139: Maybe we should at least log other errors? Everything except `ErrClosedListener` will be ignored, which seems to me like an information loss.

Danielius1922 (approved):

tcp/client/session.go
jkralik: No it's not data race:
This mutex is used when two goroutines call `SetContextValue `at the same time. Regardless of order, we want both values to be stored in the context with different keys.  Due to the fact that the operation consists of two steps, the connection can only contain one of the values without the mutex.

The operation `s.ctx.Load()` is atomic so it is goroutine safe

net/dtlslistener.go
jkralik: The `pion/udp` listener send's multiple errors like `io/timeout` `context canceled` context deadline exceeded` so they are not interesting -> we want to call accept until listen connection is closed - similar as is in TCP.

PR #417, opened 2022-12-07
Fix off-by-one error in upserts

vitprajzler:
There are two off-by-one errors in the upsert functions for message ID and message type.
`UpsertMessageId` should allow a value of 65535, while `UpsertType` should allow 255.

This fixes it, but I suspect it might break some tests or other things that rely on the existence of this bug (haven't tested). 

However, without this fix, a device sending a confirmable message to the server will recieve an ACK message back with an incorrect Message ID.

### DIFF
--- a/message/pool/message.go
+++ b/message/pool/message.go
@@ -81,7 +81,7 @@ func (r *Message) SetMessageID(mid int32) {
 
 // UpsertMessageID set value only when origin value is invalid. Only 0 to 2^16-1 values are valid.
 func (r *Message) UpsertMessageID(mid int32) {
-	if r.msg.MessageID >= 0 && r.msg.MessageID < math.MaxUint16 {
+	if r.msg.MessageID >= 0 && r.msg.MessageID <= math.MaxUint16 {
 		return
 	}
 	r.SetMessageID(mid)
@@ -99,7 +99,7 @@ func (r *Message) SetType(typ message.Type) {
 
 // UpsertType set value only when origin value is invalid. Only 0 to 2^8-1 values are valid.
 func (r *Message) UpsertType(typ message.Type) {
-	if r.msg.Type >= 0 && r.msg.Type < math.MaxUint8 {
+	if r.msg.Type >= 0 && r.msg.Type <= math.MaxUint8 {
 		return
 	}
 	r.SetType(typ)


mniestroj (commented):
I really miss more elaborate information provided in commit body. You can use information provided in PR description, e.g.:
```
There are two off-by-one errors in the upsert functions for message ID and message type.
UpsertMessageId should allow a value of 65535, while UpsertType should allow 255.

Without this fix, a device sending a confirmable message to the server will recieve an ACK message back with an incorrect Message ID.
```

While at it, please also fix `message/message.go`, which has similar check (not allowing max uint16 / uint8 values):
```
	if r.Type >= 0 && r.Type < math.MaxUint8 {
		buf = fmt.Sprintf("%s, Type: %v", buf, r.Type)
	}
	if r.MessageID >= 0 && r.MessageID < math.MaxUint16 {
		buf = fmt.Sprintf("%s, MessageID: %v", buf, r.MessageID)
	}
```

message/pool/message.go
```diff
@@ -81,7 +81,7 @@ func (r *Message) SetMessageID(mid int32) {
 
 // UpsertMessageID set value only when origin value is invalid. Only 0 to 2^16-1 values are valid.
 func (r *Message) UpsertMessageID(mid int32) {
-	if r.msg.MessageID >= 0 && r.msg.MessageID < math.MaxUint16 {
+	if r.msg.MessageID >= 0 && r.msg.MessageID <= math.MaxUint16 {
```
jkralik, line 84: Please, could you create `message.ValidateMessageID(mid int32) bool`, it will be used at both places, as asked for by @mniestroj?

message/pool/message.go
```diff
@@ -99,7 +99,7 @@ func (r *Message) SetType(typ message.Type) {
 
 // UpsertType set value only when origin value is invalid. Only 0 to 2^8-1 values are valid.
 func (r *Message) UpsertType(typ message.Type) {
-	if r.msg.Type >= 0 && r.msg.Type < math.MaxUint8 {
+	if r.msg.Type >= 0 && r.msg.Type <= math.MaxUint8 {
```
jkralik, line 102: Please, could you create `message.ValidateType(mid int32) bool`, it will be used at both places, as asked for by @mniestroj?

jkralik:
Created new PR https://github.com/plgd-dev/go-coap/pull/423 with required changes.

PR #423, opened 2023-01-04
Fix off-by-one error in upserts

jkralik:
There are two off-by-one errors in the upsert functions for message ID and message type.
UpsertMessageId should allow a value of 65535, while UpsertType should allow 255.

Without this fix, a device sending a confirmable message to the server will receive an ACK message back with an incorrect Message ID.

FYI: @mniestroj @vitprajzler

### DIFF
--- a/message/getmid.go
+++ b/message/getmid.go
@@ -3,6 +3,7 @@ package message
 import (
 	"crypto/rand"
 	"encoding/binary"
+	"math"
 	mathRand "math/rand"
 	"sync/atomic"
 	"time"
@@ -14,7 +15,7 @@ func init() {
 
 var msgID = uint32(RandMID())
 
-// GetMID generates a message id for UDP-coap
+// GetMID generates a message id for UDP. (0 <= mid <= 65535)
 func GetMID() int32 {
 	return int32(uint16(atomic.AddUint32(&msgID, 1)))
 }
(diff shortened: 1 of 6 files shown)


mniestroj (approved):
LGTM!

message/getmid.go
```diff
@@ -28,3 +29,8 @@ func RandMID() int32 {
 	}
 	return int32(uint16(binary.BigEndian.Uint32(b)))
 }
+
+// ValidateMessageID validates a message id for UDP. (0 <= mid <= 65535)
+func ValidateMessageID(mid int32) bool {
```
Danielius1922, line 34: I think `ValidateMID` would be more in line with the existing naming convention, but it's clear either way so I don't mind.

udp/coder/coder.go
```diff
@@ -46,10 +45,10 @@ func (c *Coder) Encode(m message.Message, buf []byte) (int, error) {
 	   |1 1 1 1 1 1 1 1|    Payload (if any) ...
 	   +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
 	*/
-	if m.MessageID < 0 || m.MessageID > math.MaxUint16 {
```
Danielius1922, line 49: Maybe a test of Encode with the boundary values for messageID and type?

Danielius1922 (approved):

udp/coder/coder.go
jkralik: Added

message/getmid.go
jkralik: Renamed 👍

PR #425, opened 2023-01-09
SonarCloud: lower complexity of upd/client.writeMessage

Danielius1922:


pkg/fn/funcList.go
```diff
@@ -0,0 +1,27 @@
+package fn
+
+// Slice of functions to be executed after each other
+type FuncList struct {
```
jkralik, line 4: use `type FuncList []func()` instead of struct -> you can preinitialized size then

Danielius1922: Fixed

jkralik (approved):

PR #461, opened 2023-07-21
[Fuzzing] fuzz testing support

0x34d:
using [go-fuzz](https://go.dev/security/fuzz) for fuzz testing of `go-coap`

How to run:
```bash
	go test -fuzz=FuzzDecode -fuzztime=10s $(PWD)/udp/coder/
	go test -fuzz=FuzzDecode -fuzztime=10s $(PWD)/tcp/coder/
	go test -fuzz=FuzzUnmarshalJSON -fuzztime=10s $(PWD)/message/codes/
	go test -fuzz=FuzzUnmarshal -fuzztime=10s $(PWD)/message/
```


Danielius1922:
Hello @0x34d , thank you for the contribution! 

I would like to add all of the new test cases to CI testing job. The command in our GH workflow 

```sh
go test -v -race ./... -coverpkg=./... -covermode=atomic -coverprofile=./coverage.txt
```

invokes them, but only with the preddefined seeds. It seems that running them all in a single command is not supported for now (if I read https://github.com/golang/go/issues/46312 correctly, but copying that script that invokes them all should work)

tcp/coder/coder_test.go
```diff
@@ -106,3 +106,13 @@ func BenchmarkUnmarshalMessage(b *testing.B) {
 		}
 	}
 }
+
+func FuzzDecode(f *testing.F) {
+	f.Add([]byte{211, 0, 1, 1, 2, 3, 177, 97, 1, 98, 1, 99, 1, 100, 1, 101, 16, 255, 1})
+	f.Add([]byte{211, 0, 1, 1, 2, 3, 177, 97, 1, 98, 1, 99, 1, 100, 1, 101, 16, 255, 1})
```
Danielius1922, line 112: Testing the same decode twice doesn't seem necessary, any reason to keep the second line?

udp/coder/coder_test.go
```diff
@@ -242,3 +242,16 @@ func BenchmarkUnmarshalMessage(b *testing.B) {
...
+	f.Add([]byte{0x40, 0x1, 0x30, 0x39, 0x46, 0x77, 0x65, 0x65, 0x74, 0x61, 0x67, 0xa1, 0x3, 0xff, 'h', 'i'})
+	f.Add([]byte{88, 128, 107, 170, 134, 237, 158, 132, 150, 19, 19, 159, 72, 20, 210, 14, 23, 231, 160, 183, 145, 128, 177, 14, 82, 20, 210, 255})
+	f.Add([]byte{0x48, 0x01, 0x00, 0x00, 0xB0, 0x35, 0x4C, 0xF5, 0xD9, 0x72, 0x24, 0x0D, 0x60, 0x55, 0x6C, 0x69, 0x67, 0x68, 0x74, 0x05, 0x6C, 0x69, 0x67, 0x68, 0x74})
+	f.Add([]byte{0x48, 0x45, 0x62, 0xA2, 0x8D, 0xA2, 0x29, 0x0C, 0x18, 0x0F, 0xB5, 0x4A, 0x48, 0x5E, 0x10, 0xA3, 0x88, 0x00, 0x00, 0x00, 0x00, 0x82, 0x27, 0x10, 0x52, 0x27, 0x10, 0x61, 0x16, 0xE2, 0x06, 0xDD, 0x08, 0x00, 0x42, 0x08, 0x00, 0xFF, 0x70, 0x73, 0x9F, 0xBF, 0x62, 0x65, 0x70, 0x78, 0x1A, 0x63, 0x6F, 0x61, 0x70, 0x3A, 0x2F, 0x2F, 0x31, 0x30, 0x2E, 0x31, 0x31, 0x32, 0x2E, 0x31, 0x31, 0x32, 0x2E, 0x31, 0x30, 0x3A, 0x35, 0x37, 0x39, 0x34, 0x30, 0xFF, 0xBF, 0x62, 0x65, 0x70, 0x78, 0x1E, 0x63, 0x6F, 0x61, 0x70, 0x2B, 0x74, 0x63, 0x70, 0x3A, 0x2F, 0x2F, 0x31, 0x30, 0x2E, 0x31, 0x31, 0x32, 0x2E, 0x31, 0x31, 0x32, 0x2E, 0x31, 0x30, 0x3A, 0x34, 0x36, 0x33, 0x36, 0x33, 0xFF, 0xFF, 0xFF, 0xFF})
```
Danielius1922, line 251: Do these seeds influence the generated data when the TC is invoked by 

```sh
go test -fuzz=FuzzDecode -fuzztime=10s $(PWD)/udp/coder/
```
? Other TCs have 0-2 and this has 5, seems inconsistent. Asking out of curiosity, there's no problem in keeping them all.

tcp/coder/coder_test.go
0x34d: Thanks for pointing this out.

udp/coder/coder_test.go
0x34d: Yes, the seed does influence the generation of malformed data. I removed some similar-looking seed data; now the seed dataset is looking unique.

INFO: There's a little problem with go-fuzz. It can't do seed minimization based on the path explored per seed; this is always a blind hit.

0x34d:
> Hello @0x34d , thank you for the contribution!
> 
> I would like to add all of the new test cases to CI testing job. The command in our GH workflow
> 
> ```shell
> go test -v -race ./... -coverpkg=./... -covermode=atomic -coverprofile=./coverage.txt
> ```
> 
> invokes them, but only with the preddefined seeds. It seems that running them all in a single command is not supported for now (if I read [golang/go#46312](https://github.com/golang/go/issues/46312) correctly, but copying that script that invokes them all should work)


I have a better options.

* If this project is accepted by the [oss-fuzz](https://google.github.io/oss-fuzz/) team, then [cifuzz](https://google.github.io/oss-fuzz/getting-started/continuous-integration/) is a better option.

* If they don't, then [ClusterFuzzLite](https://google.github.io/clusterfuzzlite/) is a really good option.

0x34d:
> > I have a better options.
> > 
> > * If this project is accepted by the [oss-fuzz](https://google.github.io/oss-fuzz/) team, then [cifuzz](https://google.github.io/oss-fuzz/getting-started/continuous-integration/) is a better option.
> > * If they don't, then [ClusterFuzzLite](https://google.github.io/clusterfuzzlite/) is a really good option.
> 
> Thank you for the suggestions! I'm gonna apply to `OSS-Fuzz` and see what happens.

Well, i already did this, wait for my pr on oss-fuzz.

Danielius1922:
> 1- compile_native_go_fuzzer $(pwd)/message/. FuzzUnmarshal fuzz_unmarshal `Could not find the function: func FuzzUnmarshal(f *testing.F)`

I'm looking at the [compile_native_go_fuzzer](https://github.com/google/oss-fuzz/blob/master/infra/base-images/base-builder/compile_native_go_fuzzer) script. It happens because of this line:

```sh
if [ $(grep -r "func $function" $fuzzer_filename | grep "testing.F" | wc -l) -eq 1 ]
```

when I run it locally:
```sh
...
$ grep -r "func $function" $fuzzer_filename | grep "testing.F"

/home/danielius/go/src/github.com/plgd-dev/go-coap/message/options_test.go:func FuzzUnmarshal(f *testing.F) {
/home/danielius/go/src/github.com/plgd-dev/go-coap/message/codes/codes_test.go:func FuzzUnmarshalJSON(f *testing.F) {
```

Recursive `grep` and `FuzzUnmarshal` is prefix of `FuzzUnmarshalJSON` so it always finds both.

0x34d:
> > 1- compile_native_go_fuzzer $(pwd)/message/. FuzzUnmarshal fuzz_unmarshal `Could not find the function: func FuzzUnmarshal(f *testing.F)`
> 
> I'm looking at the [compile_native_go_fuzzer](https://github.com/google/oss-fuzz/blob/master/infra/base-images/base-builder/compile_native_go_fuzzer) script. It happens because of this line:
> 
> ```shell
> if [ $(grep -r "func $function" $fuzzer_filename | grep "testing.F" | wc -l) -eq 1 ]
> ```
> 
> when I run it locally:
> 
> ```shell
> ...
> $ grep -r "func $function" $fuzzer_filename | grep "testing.F"
> 
> /home/danielius/go/src/github.com/plgd-dev/go-coap/message/options_test.go:func FuzzUnmarshal(f *testing.F) {
> /home/danielius/go/src/github.com/plgd-dev/go-coap/message/codes/codes_test.go:func FuzzUnmarshalJSON(f *testing.F) {
> ```
> 
> Recursive `grep` and `FuzzUnmarshal` is prefix of `FuzzUnmarshalJSON` so it always finds both.


Fix:
```
perl -i -pe 'BEGIN{undef $/;} s/func Benchmark.*?\(b \*testing\.B\).*?\{.*?\n\}//smg' $(pwd)/udp/coder/coder_test.go
perl -i -pe 'BEGIN{undef $/;} s/func Benchmark.*?\(b \*testing\.B\).*?\{.*?\n\}//smg' $(pwd)/tcp/coder/coder_test.go
perl -i -pe 'BEGIN{undef $/;} s/func Benchmark.*?\(b \*testing\.B\).*?\{.*?\n\}//smg' $(pwd)/message/options_test.go
sed -i 's/FuzzUnmarshal/FuzzUnmarshalData/g' $(pwd)/message/options_test.go
```

I'm working on a new PR for go-coap to remove this regex code.
Here's my PR on oss-fuzz :) https://github.com/google/oss-fuzz/pull/10746/files

PR #468, opened 2023-08-08
[Fuzzing] add cifuzz

0x34d:
Adding [CIFuzz](https://google.github.io/oss-fuzz/getting-started/continuous-integration) in workflows.


jkralik (approved):
Thx for contributing!
❤️

PR #469, opened 2023-08-08
net: Enhance observation management with ETag tracking

Danielius1922:
This commit introduces an enhancement to the observation mechanism in the network module by implementing ETag tracking. The latest ETag value associated with each observation is now stored in the internal representation.

With this update, each incoming message containing the ETag CoAP option automatically updates the ETag value for the relevant observation. When an observation is canceled, the stored ETag value is utilized. When the server detects a valid ETag match, it now responds with a VALID code and an empty payload instead of resending the content of the resource. This optimization minimizes unnecessary data transfer and reduces network load.

### DIFF
--- a/net/observation/handler.go
+++ b/net/observation/handler.go
@@ -146,10 +146,11 @@ type Observation[C Client] struct {
 	waitForResponse     atomic.Bool
 	observationHandler  *Handler[C]
 
-	private struct {
+	private struct { // members guarded by mutex
 		mutex       sync.Mutex
-		obsSequence uint32    // guarded by mutex
-		lastEvent   time.Time // guarded by mutex
+		obsSequence uint32
+		lastEvent   time.Time
+		etag        []byte
 	}
 }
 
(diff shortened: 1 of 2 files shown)


net/observation/handler.go
```diff
@@ -240,6 +251,9 @@ func (o *Observation[C]) wantBeNotified(r *pool.Message) bool {
 	if ValidSequenceNumber(o.private.obsSequence, obsSequence, o.private.lastEvent, now) {
 		o.private.obsSequence = obsSequence
 		o.private.lastEvent = now
+		if etag, err := r.ETag(); err == nil {
+			o.private.etag = etag
```
jkralik, line 255: use copy instead of assignation -> the underlayer buffer could be reused.

Danielius1922: copy() copies works based on length not capacity, so if the lengths do not match I have to allocate a new slice

jkralik (approved):

PR #471, opened 2023-08-22
Fix processing of observe notifications with ETags

Danielius1922:
Upon the commencement of the observation, the client has the ability to include ETAGs options within the GET request utilizing the observe option. Subsequently, on the server side, a Valid status code can be returned if any of the provided ETAGs are found to be a match.

### DIFF
--- a/net/blockwise/blockwise.go
+++ b/net/blockwise/blockwise.go
@@ -523,10 +523,7 @@ func isObserveResponse(msg *pool.Message) bool {
 	if err != nil {
 		return false
 	}
-	if msg.Code() == codes.Content {
-		return true
-	}
-	return false
+	return msg.Code() >= codes.Created
 }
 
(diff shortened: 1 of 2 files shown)


net/blockwise/blockwise.go
```diff
@@ -523,10 +523,8 @@ func isObserveResponse(msg *pool.Message) bool {
 	if err != nil {
 		return false
 	}
-	if msg.Code() == codes.Content {
-		return true
-	}
-	return false
+	// TODO: codes.Valid?
```
jkralik, line 526: yes also valid is fine

jkralik: Also we need to check it here:
https://github.com/plgd-dev/go-coap/blob/c1d18474cd9046c8a466465b2f5bbbd55309f410/net/blockwise/blockwise.go#L750

Issue #472, opened 2023-08-23
Support serializing and deserializing a UDP DTLS connection


oursland:
DTLS connections can have their states marshalled for storage and unmarshalled to restore the connection.  Supporting this would permit a user to maintain long-term CoAP connections without the need to for renegotiating the DTLS handshakes, improving performance particularly over unreliable and high latency networks such as LPWAN and cellular.

The underlying functionality to support this provided by the following functions from `pion/dtls`:

* [Conn.Export()](https://pkg.go.dev/github.com/pion/dtls#Conn.Export) to export the [State](https://pkg.go.dev/github.com/pion/dtls#State)
* [State.MarshallBinary()](https://pkg.go.dev/github.com/pion/dtls#State.MarshalBinary) to marshall the State to a binary object
* [State.UnmarshallBinary()](https://pkg.go.dev/github.com/pion/dtls#State.UnmarshalBinary) to unmarshall the binary to a new State object
* [Conn.Resume()](https://pkg.go.dev/github.com/pion/dtls#Resume) to resume a DTLS connection

hasheddan:
@oursland thanks for opening! This is on my radar as well 👍🏻 I recently added Connection ID support in `pion/dtls` and have a branch that should be ready by the end of this week to support in `go-coap` (#467). Are you planning to work on resuming connections or are you looking for someone else to pick it up? Happy to collaborate either way :)

PR #473, opened 2023-08-24
Add support for DTLS 1.2 Connection IDs

### LINKED ISSUE #467: Consume `pion/dtls` connection ID support
I have been working on implementing [DTLS 1.2 Connection ID](https://datatracker.ietf.org/doc/html/rfc9146) support in `pion/dtls` (https://github.com/pion/dtls/pull/558). It is close to landing, at which point I would like to start integrating the updates into `plgd-dev/go-coap` as soon as possible. This issue tracks consuming the new functionality.

hasheddan:
Updates pion/dtls dependency to bring in support for DTLS 1.2 Connection IDs.

Fixes #467


jkralik:
Nice. Is it possible to extend the example by resuming the connection?

hasheddan (commented):
@jkralik absolutely 👍🏻 will get this updated once the referenced PR in the `replace` statement is merged!

examples/dtls/cid/server/main.go
```diff
@@ -0,0 +1,92 @@
...
+}
+
+// wrappedListener wraps a net.Listener and implements a go-coap DTLS
+// server.Listener.
+// NOTE: this utility is for example purposes only. Context should be handled
+// properly in meaningful scenarios.
+type wrappedListener struct {
+	l      net.Listener
+	closed atomic.Bool
+}
```
hasheddan, line 50: @jkralik I'm not sure if there is appetite to do so, but I would be interested in `go-coap` defining a common `ContextListener` and utility wrappers to allow for the passing of `net.Listener`. For now, consumers of this functionality will need to use a custom listener as demonstrated in this example.

hasheddan (commented):
@jkralik updated with examples, as well as logs and a capture of the behavior 👍🏻

examples/dtls/cid/server/main.go
hasheddan: > But one thing - Has handshaking been moved to Conn.Read/Write from listener.Accept? If not, we need to incorporate it with https://github.com/plgd-dev/go-coap/blob/master/net/dtlslistener.go#L37.

@jkralik I'm not sure I'm following your suggestion here. Handshaking has not moved -- from the perspective of the existing `DTLSListener` nothing has changed. In fact, setting the `connectionIDGenerator` on the `pion/dtls` config passed to `NewDTLSListener` would enable connection ID support. The issue is that the underlying `pion/transport/udp` package [being used](https://github.com/plgd-dev/go-coap/blob/12f9eb2affdaa01ec1a7c64cef4b3dad41f39241/net/dtlslistener.go#L95) does not support routing based on Connection ID. A replacement has been introduced in https://github.com/pion/dtls/tree/master/internal/net/udp, which is in `internal` for now so that it can be iterated upon prior to folks taking a direct dependency on its API. It can be consumed by using the [`pion/dtls.Listen` function](https://github.com/pion/dtls/blob/609e5bee6eb0269574fed9f5d50127fdb360b846/listener.go#L16).

I do think it would be good for the functionality to become the default in `go-coap`, but right now the `DTLSListener` is restrictive about the behavior it allows (anecdotally, we use a different implementation of the `DTLSListener` internally so introduce additional functionality). This PR maintains the existing `go-coap` functionality, while also demonstrating how a consumer can provide their own listener for custom functionality (including connection IDs).

It seems like previously the `pion/dtls.Listener` was [used](https://github.com/plgd-dev/go-coap/pull/379/files#diff-aaa722ddf89aaefa433bb78b18f5e8f69544665651b12bfa78d1e6dc57a2f1a6L42), but then was moved away from in order to introduce parallel handshakes. My understanding of this functionality is that the `DTLSListener` essentially just continuously accepts connections so that the handshake can begin and complete, perhaps before the DTLS `Server` calls `AcceptWithContext()`. It seems the major blocker from using the `pion/dtls.Listener` is the [starting of the `pion/dtls.Server` [in a go pool](https://github.com/plgd-dev/go-coap/blob/12f9eb2affdaa01ec1a7c64cef4b3dad41f39241/net/dtlslistener.go#L123). My question is whether this could be functionality introduced in `pion/dtls` instead (i.e. stop blocking `Accept()` on completion of the handshake), which would allow `go-coap` to treat the DTLS listener as just a `net.Listener` (with the generic wrapping to support context handling).

hasheddan: (Note: this is what I would like to do long term, but the changes in this PR enable folks to implement custom behavior to enable connection ID support, while keeping the previous behavior of `go-coap` unchanged when using the provided `DTLSListener` implementation. This feels like a good intermediate step to me as it does not introduce any breaking changes in `go-coap`, but I am open to discussion and happy to explore other avenues as well!)

go.mod
```diff
@@ -5,8 +5,8 @@ go 1.18
 require (
```
hasheddan, line 5: Looks like linter is stuck on `replace` statement, but it has been removed and required functionality has been merged upstream 👍🏻

examples/dtls/cid/server/main.go
jkralik: > It seems like previously the pion/dtls.Listener was [used](https://github.com/plgd-dev/go-coap/pull/379/files#diff-aaa722ddf89aaefa433bb78b18f5e8f69544665651b12bfa78d1e6dc57a2f1a6L42), but then was moved away from in order to introduce parallel handshakes. My understanding of this functionality is that the DTLSListener essentially just continuously accepts connections so that the handshake can begin and complete, perhaps before the DTLS Server calls AcceptWithContext(). It seems the major blocker from using the pion/dtls.Listener is the [starting of the pion/dtls.Server [in a go pool](https://github.com/plgd-dev/go-coap/blob/12f9eb2affdaa01ec1a7c64cef4b3dad41f39241/net/dtlslistener.go#L123). My question is whether this could be functionality introduced in pion/dtls instead (i.e. stop blocking Accept() on completion of the handshake), which would allow go-coap to treat the DTLS listener as just a net.Listener (with the generic wrapping to support context handling).

There is a discussion (https://github.com/pion/dtls/issues/279) about moving the handshake, as it is in the `golang/crypto` package, to the `Read/Write` functions.

I agree with you; we can proceed step by step.

jkralik (approved):

jkralik:
Thx for contribution :).

jkralik:
resolves #472

nicgirault:
@jkralik your last comment mark #472 as resolved but the issue is still opened. Is serializing and deserializing a UDP DTLS connection functional?

PR #484, opened 2023-09-26
Include whitespace between semicolon and encoding

hasheddan:
Updates the text/plain media type string to use a space between the semicolon and the encoding. While this is not strictly required, it more closely aligns with the media types in the content format registry.

https://www.iana.org/assignments/core-parameters/core-parameters.xhtml

### DIFF
--- a/message/option.go
+++ b/message/option.go
@@ -161,7 +161,7 @@ type MediaType uint16
 
 // Content formats.
 var (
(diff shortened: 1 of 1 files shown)


jkralik (approved):

jkralik:
Thx for contribution :)

PR #497, opened 2023-10-18
Clone router map before returning it to user

HRogge:
mux/Router.GetRoutes() leaks access to the (lock-protected) internal map of the router. Clone the map before returning it to the user.

### DIFF
--- a/mux/router.go
+++ b/mux/router.go
@@ -9,8 +9,7 @@ import (
 	"github.com/plgd-dev/go-coap/v3/message"
 	"github.com/plgd-dev/go-coap/v3/message/codes"
 	"github.com/plgd-dev/go-coap/v3/message/pool"
-	// TODO: switch to non-exp maps package when dependency hits 1.20
-	"golang.org/x/exp/maps"
+	"golang.org/x/exp/maps" // TODO: switch to non-exp maps package when dependency hits 1.20
 )
 
 type ResponseWriter = interface {
--- a/mux/router.go
+++ b/mux/router.go
@@ -6,12 +6,11 @@ import (
 	"io"
 	"sync"
 
-	// TODO: switch to non-exp maps package when dependency hits 1.20
-	"golang.org/x/exp/maps"
-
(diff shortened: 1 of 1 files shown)


mux/router.go
```diff
@@ -6,6 +6,9 @@ import (
 	"io"
 	"sync"
 
+	// TODO: switch to non-exp maps package when dependency hits 1.20
+	"golang.org/x/exp/maps"
```
Danielius1922, line 10: Put this import after `"github.com/plgd-dev/go-coap/v3/message/pool"`, the linter wants alphabetical order.

Danielius1922: I think we can change the signatures of `GetRoute` and `GetRoutes`, because `Route` has no exported members, the only thing a caller can do is call the single exported method `GetRouteRegexp`. So instead of returning `*Route` and map[string]*Route, we can return `string` and `map[string]string` and the string is the representation returned by `GetRouteRegexp` for given *Route. 
`Route`  `GetRouteRegexp` can be both unexported.

@jkralik

HRogge: Yes, unexporting would change the solution for the mux, but it would break all external code depending on this function.

jkralik: I agree with @HRogge . Pls could you apply this patch to fix linter ?:
```
diff --git a/mux/router.go b/mux/router.go
index 92f5e2b..cb61e1f 100644
--- a/mux/router.go
+++ b/mux/router.go
@@ -9,8 +9,7 @@ import (
        "github.com/plgd-dev/go-coap/v3/message/pool"
-       // TODO: switch to non-exp maps package when dependency hits 1.20
-       "golang.org/x/exp/maps"
+       "golang.org/x/exp/maps" // TODO: switch to non-exp maps package when dependency hits 1.20
 )
 
 type ResponseWriter = interface {
 ```

jkralik (approved):

jkralik:
Thx for contribution :).

PR #498, opened 2023-10-19
mux: filter for invalid root path pattern

HRogge:
Mux router has trouble with root path matching because of the different semantics of empty path and "/" path. Add the already existing filter into all relevant router functions and add extend test cases to check all combinations.

### DIFF
--- a/mux/router_test.go
+++ b/mux/router_test.go
@@ -5,6 +5,9 @@ import (
 
 	"github.com/plgd-dev/go-coap/v3/mux"
 	"github.com/stretchr/testify/require"
+
+	// TODO: replace with standard maps package as soon as Go dependency hits 1.20
+	"golang.org/x/exp/maps"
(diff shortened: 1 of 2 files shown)


mux/router_test.go
```diff
@@ -5,6 +5,9 @@ import (
 
 	"github.com/plgd-dev/go-coap/v3/mux"
 	"github.com/stretchr/testify/require"
+
+	// TODO: replace with standard maps package as soon as Go dependency hits 1.20
+	"golang.org/x/exp/maps"
```
jkralik, line 10: Please remove the empty line, reorder the imports if needed, and add a comment on the same line as the import to satisfy the linter.

jkralik (approved):
Just fix the import :)

jkralik:
Thx for contribution :).

PR #499, opened 2023-10-25
fix: ensure content type strings match IANA registry

### LINKED ISSUE #492: Media type strings do not match IANA registry
Similar to #484, many of the [media types](https://github.com/plgd-dev/go-coap/blob/24bc8fe40ed44e700eaec7bd8969cfc0f102a96f/message/option.go#L193) do not match their representation in the [IANA registry](https://www.iana.org/assignments/core-parameters/core-parameters.xhtml#content-formats), which can make it difficult to handle them generically. Most of the offenders are due to the inclusion of the RFC suffix.

hasheddan:
Updates string representation of content-types to match the definitions in the IANA registry.

https://www.iana.org/assignments/core-parameters/core-parameters.xhtml#content-formats

Fixes #492


jkralik (approved):
Thx for the fix :+1:

Issue #500, opened 2023-10-27
Ping (CON Empty Message) Notifications


jeffwelder-ellenbytech:
I would like my application to get notified of Confirmable Empty Message requests (Pings) to keep track of routable paths during observe.

Currently I believe this code prevents that from happening: 
https://github.com/plgd-dev/go-coap/blob/a48812ee4db5dbec9c8fb43496e8fc83dee79f7b/udp/client/conn.go#L785C2-L787C3
``` go
if r.Code() == codes.Empty && r.Type() == message.Confirmable && len(r.Token()) == 0 && len(r.Options()) == 0 && r.Body() == nil {
    cc.ProcessReceivedMessageWithHandler(r, cc.handlePong)
    return true
}
```

I would like to be able to handle this route either through `router.DefaultHandle()`, or passing in a configuration that allows you to create your own `handlePong` function.

Are there any concerns to bellying this up to the application?

jeffwelder-ellenbytech:
Closed by #511

PR #501, opened 2023-10-27
feat: Propagate CoAP Ping (CON Empty) to Handlers

jeffwelder-ellenbytech:
Allow ping request to be propagated to application handlers to allow application to keep track of devices that are pinging.
This is useful when you have devices observing resources that have to ping to keep routes open.

See issue #500


udp/client/conn.go
```diff
@@ -744,8 +744,9 @@ func (cc *Conn) handlePong(w *responsewriter.ResponseWriter[*Conn], r *pool.Mess
 func (cc *Conn) handleSpecialMessages(r *pool.Message) bool {
 	// ping request
 	if r.Code() == codes.Empty && r.Type() == message.Confirmable && len(r.Token()) == 0 && len(r.Options()) == 0 && r.Body() == nil {
+		// handle ping request by sending pong, but still propogate the message
+		// to any handlers incase the application wants to know when pings happened
 		cc.ProcessReceivedMessageWithHandler(r, cc.handlePong)
```
jkralik, line 749: You shouldn't propagate the ping to the handler via `cc.handleSpecialMessages.C()` because the response has already been processed. Instead, you can extend the options related to the callback/channel that will be triggered after the request has been [unmarshalled line 782](https://github.com/plgd-dev/go-coap/blob/a48812ee4db5dbec9c8fb43496e8fc83dee79f7b/udp/client/conn.go#L782) in **the goroutine**.

udp/client/conn.go
```diff
@@ -782,9 +783,7 @@ func (cc *Conn) Process(datagram []byte) error {
 	req.SetSequence(cc.Sequence())
 	cc.checkMyMessageID(req)
 	cc.inactivityMonitor.Notify()
-	if cc.handleSpecialMessages(req) {
-		return nil
-	}
+	cc.handleSpecialMessages(req)
```
jkralik, line 786: Pls return back. Special messages are processed by package.

jkralik (commented):
Pls also extend TCP by same way. We want to provide similar API for both protocols.

jeffwelder-ellenbytech:
@jkralik I reverted the changes to allow the package to process the Pings and attempted to extend an options through a `WithPingMonitor()` function.

This option allows you to get a con object whenever a device pings the server.
I modeled the option somewhat after the inactivityMonitor, and tried to extended it to udp/tcp/dtls.

Please let me know if there's a better way to do this. :)

udp/client/conn.go
```diff
@@ -741,12 +745,21 @@ func (cc *Conn) handlePong(w *responsewriter.ResponseWriter[*Conn], r *pool.Mess
 	cc.sendPong(w, r)
 }
 
+func (cc *Conn) isPingMessage(r *pool.Message) bool {
+	if r.Code() == codes.Empty && r.Type() == message.Confirmable && len(r.Token()) == 0 && len(r.Options()) == 0 && r.Body() == nil {
```
Danielius1922, line 749: Can be simplified to 

```
return r.Code() == codes.Empty && r.Type() == message.Confirmable && len(r.Token()) == 0 && len(r.Options()) == 0 && r.Body() == nil
```

Danielius1922:
> @jkralik I reverted the changes to allow the package to process the Pings and attempted to extend an option through a `WithPingMonitor()` function.
> 
> This option allows you to get a con object whenever a device pings the server. I modeled the option somewhat after the inactivityMonitor, and tried to extended it to udp/tcp/dtls.
> 
> Please let me know if there's a better way to do this. :)

This implementation is fine, however we talked about this with @jkralik today and we would prefer to change this from just a ping request monitor to a general request monitor to have a more general purpose feature. Any arguments why that could be problematic? If not then, please:

- [ ] change `PingMonitorFunc` to `RequestMonitorFunc` which takes two arguments (first is the connection as before, and extend it by a second argument which will be the request message `pool.Message`)
- [ ] change isPingMessage to IsPingMessage so it gets exported and the library user can easily implement the ping monitor functionality

tcp/client/conn.go
```diff
@@ -59,6 +60,7 @@ func NewConn(
 	connection *coapNet.Conn,
 	createBlockWise func(cc *Conn) *blockwise.BlockWise[*Conn],
 	inactivityMonitor InactivityMonitor,
+	requestMonitor RequestMonitorFunc,
```
Danielius1922, line 63: This breaks the signature of the public API, we should avoid that in non major library updates. (Same with `NewSession` in `tcp/client/session.go`  and `NewConn` in `udp/client/conn.go`).

The signatures of the NewConn/NewSession should remain the same as before, but we need new functions that take requestMonitor argument.

@jkralik do we prefer something like:

```go
func NewConnWithRequestMonitor(connection *coapNet.Conn, createBlockWise func(cc *Conn) *blockwise.BlockWise[*Conn], inactivityMonitor InactivityMonitor, requestMonitor RequestMonitorFunc, cfg *Config)
{
...
}
```

or perhaps we should go for function with variadic number of options?

```go
type Option struct {
	CreateBlockWise func(cc *Conn) *blockwise.BlockWise[*Conn]
	InactivityMonitor InactivityMonitor
        RequestMonitor RequestMonitorFunc
}

func NewConnWithOpts(connection *coapNet.Conn, cfg *Config, opts ...Option)
{
...
}
```

I would prefer the variadic version.

jkralik:
@jeffwelder-ellenbytech Pls could you update the PR and fix the issue mentioned in https://github.com/plgd-dev/go-coap/pull/501#discussion_r1383503732

jeffwelder-ellenbytech:
> @jeffwelder-ellenbytech Pls could you update the PR and fix the issue mentioned in [#501 (comment)](https://github.com/plgd-dev/go-coap/pull/501#discussion_r1383503732)

Sure, using the `func NewConnWithOpts(connection *coapNet.Conn, cfg *Config, opts ...Option)`?

jeffwelder-ellenbytech:
> @jeffwelder-ellenbytech Pls could you update the PR and fix the issue mentioned in [#501 (comment)](https://github.com/plgd-dev/go-coap/pull/501#discussion_r1383503732)

@jkralik Sure, using the func NewConnWithOpts(connection *coapNet.Conn, cfg *Config, opts ...Option)?

PR #518, opened 2023-12-05
update pion/dtls to v2.2.8

shimataro:
[The v2 branch of pion/dtls no longer contains "pkg/net"](https://github.com/pion/dtls/tree/v2/pkg) and [the `Client()` / `Server()` functions have been changed to accept `net.Conn`](https://github.com/pion/dtls/blob/v2/conn.go).

I think this PR allows go-coap to update pion/dtls to v2.2.8...maybe.

(I am not familiar with the use of pion/dtls, so sorry if there are any mistakes.)


Danielius1922:
Hi @shimataro it seems you have an older version of the repository in your fork, I've updated the dependencies on Dec 1, if you check the latest go.mod from the master branch you will see that pion/dtls and pion/transport have been updated to recent versions:

https://github.com/plgd-dev/go-coap/blob/b5adef4fb2ee0672ce21fe32726f841d9f904b9c/go.mod#L8-L9

Danielius1922:
Closing after month of inactivity, reopen if needed.

PR #523, opened 2024-01-08
Upgrade dependencies

Danielius1922:


go.mod
```diff
@@ -5,21 +5,24 @@ go 1.18
...
 	github.com/pmezard/go-difflib v1.0.0 // indirect
-	golang.org/x/crypto v0.16.0 // indirect
-	golang.org/x/sys v0.15.0 // indirect
+	golang.org/x/crypto v0.17.0 // indirect
+	golang.org/x/sys v0.16.0 // indirect
 	gopkg.in/yaml.v3 v3.0.1 // indirect
 )
+
+// note: github.com/pion/dtls/v2/pkg/net package is not yet available in release branches
+exclude github.com/pion/dtls/v2 v2.2.9
```
Danielius1922, line 28: @hasheddan Hey there, do you know if pkg/net package will be added to release branches? It's available in the master branch, but not in the release branches (see v2.2.9 tag for example), so I have to exclude it here so the upgrade bot doesn't go for v2.2.9 and break compilation.

jkralik (approved):

Issue #526, opened 2024-01-23
ListenAndServeWithOptions doesn't correctly start UDP server


sdeboeck:
The ListenAndServeWithOptions function does not correctly start the UDP server when using udpServer.Option. 
The UDP options are not being populated, leading to unexpected behavior.

**Steps to Reproduce**

1. Use the ListenAndServeWithOptions function to start a CoAP server with UDP
2. Provide options.WithMux option.
3. The server doesn't invoke the desired handlers.

https://github.com/plgd-dev/go-coap/blob/c61c6485126a661ad05a5b49080ed604e7ad3cc3/server.go#L87

Danielius1922:
Fixed by the linked PR.

PR #527, opened 2024-01-23
Fix UDP Server Option Duplication in ListenAndServeWithOptions

sdeboeck:
This change ensures that when a UDP option is provided in the options list, it is appropriately added to both the TCP and UDP option slices, ensuring consistent behavior for both server types.

#526


jkralik (approved):
Thx for your contribution. 👍

PR #554, opened 2024-05-15
Blockwise: Use hash instead of token for cache

### LINKED ISSUE #512: Blockwise requests should be correlated based on message options
Hello,

From what I understand from the current go-coap code, individual requests of the same block wise operation are identified based on an identical token of the individual requests. Unfortunately this is not correct per CoAP specification and go-coap only works with block wise with libraries that also have this peculiarity.

Instead of relying on the token, go-coap should rely on identical options between the individual requests, except for the NoCacheKey options and the block1/block2 options. See also the definition of "matchable" in [rfc9175](https://datatracker.ietf.org/doc/html/rfc9175#name-terminology) that clarifies this.

Best regards,

mpenate-ellenbytech:
Credit to @zworks-okada for initial work regarding rx transfers. Expanded to include tx.

Closes #512

Needs more thorough testing than what I was able to do. Willing to help address concerns.


mpenate-ellenbytech:
Fixed linting issues.

I should add that I am not a golang programmer, and addressing test failures will be difficult without any help. I can ask a colleague for some of their input but direct help would be appreciated.

net/blockwise/blockwise.go
```diff
@@ -504,7 +511,8 @@ func (b *BlockWise[C]) continueSendingMessage(w *responsewriter.ResponseWriter[C
 	}
 	var sendMessage *pool.Message
 	var more bool
-	b.sendingMessagesCache.LoadWithFunc(r.Token().Hash(), func(value *cache.Element[*pool.Message]) *cache.Element[*pool.Message] {
+	matchableHash := generateMatchableHash(r.Options(), w.Conn().RemoteAddr(), r.Code())
```
jkralik, line 514: @mpenate-ellenbytech Thank you for contribution :)

This is an issue: the hash for the response will not match the request hash because r.Code() is different, and r.Options(), such as the URI, are not part of the response. Therefore, you cannot pair it correctly for block1.

For sending a big payload (Block1):
You can use RemoteAddress and MessageID, which need to be registered for each new request of type Confirmation/NonConfirmation and response. If MessageID is not supported, then only the remote address can be used to verify block1.

For receiving a big payload (Block2):
When you want to locate the sent request, this will not work because the hash will again be different. Therefore, you need to pair it with the request using MessageID. However, you need to set a new MessageID for each exchange (request and response).

BTW: For blockwise transfer via TCP, the token must be used as implemented now.

Therefore, create blockwiseTCP.go (original code) and blockwiseUDP.go, where blockwiseUDP.go will utilize MessageIDs.

net/blockwise/blockwise.go
```diff
@@ -732,6 +741,38 @@ func (b *BlockWise[C]) getCachedReceivedMessage(mg *messageGuard, r *pool.Messag
...
+This function concatenates the IDs and values of relevant options, the string representation of the remote address,
+and the code of the message to generate a hash that can be used to match requests.
+*/
+func generateMatchableHash(options message.Options, remoteAddr net.Addr, code codes.Code) uint64 {
+	input := make([]byte, 0, 512)
+
+	for _, opt := range options {
+		switch opt.ID {
+		// Skip Blockwise Options and NoCacheKey Options
+		case message.Block1, message.Block2, message.Size1, message.Size2, message.RequestTag:
```
chrysn, line 762: I'm not sure what this hash is used for, but you may want to exclude RequestTag here.

Yes, the text says it is not part of the matchable set, but later on it says (roughly) that even if two requests are matchable, if they have different request tags, they should not be matched.

When I wrote RFC9175, we had the choice of wording while not changing the actual mechanism: We went for easier for those implementing the client side (where Request-Tag is excluded for the definition of being matchable, and then if two requests are matchable even if you don't want them to be, you add a Request-Tag and there is an extra rule on it for the server), or easier for those implementing the server side (when Request-Tag would not have been special, but in all the client sections we would have said "if the request are matchable except for possibly differing Request-Tag values"). We went for the former -- I'm not sure it was the right decision, especially if generateMatchableHash is used on the server side. (I can't tell easily without having more context of go-coap, was just pointed to this issue by a colleague).

chrysn: Also, the NoCacheKey options could more generally be recognzied from their option properties (lower bits of the number) rather than enumerating them (which will not be exhaustive).

PR #560, opened 2024-06-07
Blockwise progress

### LINKED ISSUE #236: Blockwise progress
https://github.com/plgd-dev/go-coap/blob/ab9c0385aa13d979fc010b73e4997b31541073c8/net/blockwise/blockwise.go#L533-L536

Is there any reason in the above code, the next() call cannot be called first? Since you remove the Block2 option prior to calling next(), it prevents the handler from using r.Options.GetUint32(message.Block2) in order to track the progress (such as compute percent complete or store progress)

Is there some other way to accomplish this that I have missed?

krbvroc1:
Do not remove the block until AFTER the next() call so that handler can access the Block2 option. This allows the handle to keep track of the transfer. For example, keeping track of Block X of Y, or report status every 50 blocks, or on last block signal a completion event

Closes #236


jkralik (approved):

jkralik:
Thx for your contribution. 👍

PR #589, opened 2024-11-04
[udp] make response message cache configurable and update default

### LINKED ISSUE #586: Responses are unconditionally cached for `EXCHANGE_LIFETIME`
When processing a response, all responses to `CON` and `NON` messages are cached for `247` seconds (`EXCHANGE_LIFETIME`):

https://github.com/plgd-dev/go-coap/blob/91a04eabf7408addb30cca17ea86c07092d9e102/udp/client/conn.go#L721

https://github.com/plgd-dev/go-coap/blob/91a04eabf7408addb30cca17ea86c07092d9e102/udp/client/conn.go#L623

https://github.com/plgd-dev/go-coap/blob/91a04eabf7408addb30cca17ea86c07092d9e102/udp/client/conn.go#L34

This is due to the language in [Section 4.5 of RFC 7252](https://datatracker.ietf.org/doc/html/rfc7252#section-4.5).

> A recipient might receive the same Confirmable message (as indicated
   by the Message ID and source endpoint) multiple times within the
   EXCHANGE_LIFETIME ([Section 4.8.2](https://datatracker.ietf.org/doc/html/rfc7252#section-4.8.2)), for example, when its
   Acknowledgement went missing or didn't reach the original sender
   before the first timeout.  The recipient SHOULD acknowledge each
   duplicate copy of a Confirmable message using the same
   Acknowledgement or Reset message but SHOULD process any request or
   response in the message only once.

Each `Conn` maintains its own `responseMsgCache`:

https://github.com/plgd-dev/go-coap/blob/91a04eabf7408addb30cca17ea86c07092d9e102/udp/client/conn.go#L148

The cache is checked for expiration whenever the `Conn` expirations are checked:

https://github.com/plgd-dev/go-coap/blob/91a04eabf7408addb30cca17ea86c07092d9e102/udp/client/conn.go#L910

This allows for a recipient to respond to duplicate requests without processing multiple times. However, it also means that endpoints that send large volumes of data have the potential to grow very large caches. There is no mechanism in `go-coap` today to modify the `EXCHANGE_LIFETIME`, and the `responseMsgCache` has unbounded size. This is particularly an issue in the case of blockwise transfers, where a single large object may be transmitted over many messages. With the current implementation, that entire large object could end up in the cache. Furthermore, if multiple connections are all fetching the same large object, many copies of the same object may end up in the cache (i.e. one in each `Conn` `responseMsgCache`). In the same section of RFC 7252, it allows for relaxation of caching behavior:

> A server might relax the requirement to answer all retransmissions
      of an idempotent request with the same response ([Section 4.2](https://datatracker.ietf.org/doc/html/rfc7252#section-4.2)), so
      that it does not have to maintain state for Message IDs.  For
      example, an implementation might want to process duplicate
      transmissions of a GET, PUT, or DELETE request as separate
      requests if the effort incurred by duplicate processing is less
      expensive than keeping track of previous responses would be.

There are a few different mechanisms that could be introduced (potentially alongside each other) to address this behavior:
- Make a `Conn` `responseMsgCache` size bounded such that it cannot grow unconditionally. Using an LRU strategy could be simple approach to ensuring that responses sent most recently are more likely to be in the cache.
- Support specifying messages that are not to be cached.
- Support configuring the `EXCHANGE_LIFTETIME`.

I would be happy to work on implementation if these features are of interest. Thanks!

hasheddan:
Modifies UDP conn implementation to allow for a custom response message cache to be provided. This change is implemented in such a way that API compatibility with previous v3 releases is preserved. Users may now provide an option to supply their own cache implementation. Structured messages, rather than their serialized representation, are now passed to the cache implementation to allow for caching decisions to be made in the cache implementation. For example, it may be desirable to skip caching blockwise message responses if the entire underlying data being transferred is also cached. The cache implementation is responsible for cloning messages or otherwise ensuring that it is not storing data that may subsequently be modified.

A slight functional change is also included in this update in that the default cache implementation now has pseudo-LRU functionality via ccache. When 30 newer messages are added to the cache, the 31st is marked for deletion. Note that deletion is not immediate in ccache and expired items can be returned. A check is added to ignore items that are returned expired.

Further performance improvements can be made in the default cache implementation. For example, while this implementation dramatically improves memory consumption for connections performing large volumes of requests, it also increases the total overhead of every new connection. However, because the implementation is now pluggable, improvements can be made without requiring updates to the library or breaking API changes.

Fixes #586 

> Note: keeping in draft for discussion at the moment.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -4,6 +4,7 @@ go 1.20
 
 require (
 	github.com/dsnet/golib/memfile v1.0.0
+	github.com/karlseguin/ccache/v3 v3.0.6
 	github.com/pion/dtls/v3 v3.0.2
 	github.com/stretchr/testify v1.9.0
 	go.uber.org/atomic v1.11.0
--- a/go.sum
+++ b/go.sum
@@ -2,6 +2,8 @@ github.com/davecgh/go-spew v1.1.1 h1:vj9j/u1bqnvCEfJOwUhtlOARqs3+rkHYY13jYWTU97c
 github.com/davecgh/go-spew v1.1.1/go.mod h1:J7Y8YcW2NihsgmVo/mv3lAwl/skON4iLHjSsI+c5H38=
 github.com/dsnet/golib/memfile v1.0.0 h1:J9pUspY2bDCbF9o+YGwcf3uG6MdyITfh/Fk3/CaEiFs=
 github.com/dsnet/golib/memfile v1.0.0/go.mod h1:tXGNW9q3RwvWt1VV2qrRKlSSz0npnh12yftCSCy2T64=
+github.com/karlseguin/ccache/v3 v3.0.6 h1:6wC04CXSdptebuSUBgsQixNrrRMUdimtwmjlJUpCf/4=
+github.com/karlseguin/ccache/v3 v3.0.6/go.mod h1:b0qfdUOHl4vJgKFQN41paXIdBb3acAtyX2uWrBAZs1w=
 github.com/pion/dtls/v3 v3.0.2 h1:425DEeJ/jfuTTghhUDW0GtYZYIwwMtnKKJNMcWccTX0=
 github.com/pion/dtls/v3 v3.0.2/go.mod h1:dfIXcFkKoujDQ+jtd8M6RgqKK3DuaUilm3YatAbGp5k=
 github.com/pion/logging v0.2.2 h1:M9+AIj/+pxNsDfAT64+MAVgJO0rsyLnoJKCqf//DoeY=
--- a/udp/client/conn.go
+++ b/udp/client/conn.go
@@ -6,9 +6,11 @@ import (
 	"fmt"
 	"math"
 	"net"
+	"strconv"
 	"sync"
 	"time"
 
+	"github.com/karlseguin/ccache/v3"
 	"github.com/plgd-dev/go-coap/v3/message"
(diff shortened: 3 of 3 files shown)


hasheddan (commented):
I have opened https://github.com/plgd-dev/go-coap/pull/590 as a replacement for this PR. I've played around with a number of popular Go in-memory caches, and though I do think the default cache used should employ some LRU functionality to keep the size bounded, all implementations I have tested have exhibited one or more of the following drawbacks:
- They don't also support TTL expiration.
- They are designed to run forever and do not expose functionality for stopping their background worker process (https://github.com/hashicorp/golang-lru/issues/159).
- There are questions regarding cleaning up the cache after use (https://github.com/karlseguin/ccache/issues/3, https://github.com/karlseguin/ccache/issues/59, https://github.com/karlseguin/ccache/issues/81).
- They require cleanup at all. Initially I liked the property of the cache expirations being divorced from the `CheckExpirations()` functionality of the `udp.Conn`. However, this mandates that `Close()` be called, or `Run()` be called and eventually exit. If a `udp.Conn` is created and then not manually cleaned up, resources could be leaked.

I think we could add minimal LRU functionality to the existing cache, though for now #590 makes it such that caching decisions can be owned by consumers, which allows for providing an LRU cache if desired. I'd prefer to move forward with that strategy and see if we can determine a better default implementation based on feedback.

hasheddan:
Replaced by #590

PR #619, opened 2025-09-11
Feature/fix blockwise requires message

Theoyor:
## Summary by CodeRabbit

- New Features
  - Improved reliability of blockwise transfers with exchange-based keying, enhancing concurrency and handling when tokens are absent.

- Refactor
  - Updated blockwise APIs to require a remote address parameter; integrated across TCP/UDP clients. Adjust your calls to pass the connection’s RemoteAddr().

- Chores
  - Added IDE configuration and VCS mapping files.
  - Expanded IDE/git ignore rules.
  - Updated example client to use UDP port 5685 by default.


jkralik:
> This PR is draft until:
> 
> * I got feedback from an active contributor @jkralik wrote most of the blockwise code
> * I tested multiple scenarios including coap observe, handling of multiple blockwise transfers to different resources without tokens
> * I adjusted the unit tests to be correctly interacting with the proposed fix.
> 
> I would greatly appreciate feedback!

Hi @Theoyor, thank you for your contribution. Here are my two cents on what I think needs to be done. Of course, this should also be covered by tests.

If the token is nil, only the remote address should be used as the exchange key. According to the CoAP blockwise transfer specification:

**Token usage (RFC 7252 §5.3) with blockwise**  
  - The **Token** binds a request to its response.  
  - If the Token is **non-zero**, multiple parallel exchanges with the same endpoint are possible, since each exchange can be disambiguated by `(endpoint, token)`.   However, **per resource only one blockwise transfer may be active at a time**. 
  - If the Token is **zero (nil)**, there is no disambiguation. In that case, only **one transfer per endpoint** can be active at a time. [RFC 7252 §5.3](https://datatracker.ietf.org/doc/html/rfc7252#section-5.3)

**Summary:**  
- **Nil Token (0):** at most **one transfer per endpoint**.  In this case, if a nil token is used, the client must be limited to sending only one request–response exchange at a time. The client must wait until the previous exchange completes before sending a new request.
- **Non-nil Token:** multiple transfers are possible, but **only one per resource**, disambiguated by Token.


**Example with Nil Token**
```
CLIENT                                                     SERVER
     |                                                              |
     | CON [MID=1234], POST /soap, Block1 NUM=0, M=1, Size=128 --> |
     |                                                              |
     | <------ ACK [MID=1234], 2.31 Continue, Block1 NUM=0, M=1    |
     |                                                              |
     | CON [MID=1235], POST /soap, Block1 NUM=1, M=1, Size=128 --> |
     |                                                              |
     | <------ ACK [MID=1235], 2.31 Continue, Block1 NUM=1, M=1    |
     |                                                              |
     | CON [MID=1236], POST /soap, Block1 NUM=2, M=0, Size=128 --> |
     |                                                              |
     | <------ ACK [MID=1236], Empty                                 |
     | <------ CON [MID=5000], 2.04 Changed, Block2 NUM=0, M=1      | <--- problem with disambiguation, only remote address could be used at CLIENT
     |                                                              |
     | ACK [MID=5000], POST /soap, Block2 NUM=0, M=1 -------------->|
     |                                                              |
     | <------ CON [MID=5001], 2.04 Changed, Block2 NUM=1, M=0      |
     |                                                              |
     | ACK [MID=5001], POST /soap, Block2 NUM=1, M=0 -------------->|
     |                                                              |
     ... (repeat for remaining Block2) ...
```

Theoyor:
Hi @jkralik,

thanks for your review. Did you spot any issues with the code itself apart from the topic, you mentioned (and the tests)?

Regarding your comment, I copuldn't find the exact information you mantioned at RFC 7257 §5.3, however it says:

"The client SHOULD generate tokens in such a way that tokens currently in use for a given **source/destination endpoint** pair are unique.  [...] An empty token value is appropriate e.g., when **no other tokens are in use to a destination**, [...]"

From which I infer that your statement is still valid and I will change the implementation accordingly.

PR #626, opened 2025-10-08
Feature/fix blockwise requires message

jkralik:
## Summary by CodeRabbit

- New Features
  - Block-wise transfers now track exchanges per remote address for more reliable per-peer handling.

- Refactor
  - Block-wise APIs and internal flows updated to propagate remote address; TCP and UDP clients adjusted accordingly.

- Tests
  - Added UDP libcoap-based blockwise tests (GET/PUT/POST) with large payloads and test utilities.

- Chores
  - Examples switched to UDP port 5685.
  - .gitignore adjusted to ignore only the repository root IDE directory (/.idea/).

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -131,13 +131,9 @@ jobs:
 
       - name: Install coap-client
         run: |
-          if [ "${{ matrix.os }}" == "ubuntu-latest" ]; then
             sudo apt-get update
             sudo apt-get install -y libcoap3-bin
             sudo ln -s /usr/bin/coap-client-notls /usr/bin/coap-client 
-          elif [ "${{ matrix.os }}" == "macOS-latest" ]; then
-            brew install libcoap
-          fi
         shell: bash
 
       - name: Checkout


Theoyor:
I created #627 for unit tests please take a look. It merges into this branch so you can make changes before it is merged to master

Also one would need to add coap-client into the test environment, which I cannot do.

Theoyor:
Hi @jkralik thanks for installing and merging. I was completely condused on why it doesn't work anymore. It did half a month ago and now it also doesn't run on my system anymore. Now I found the issue and I am really unsure what to do exactly:

I always tested with a pre-compiled libcoap coap-client which I now found out is quite outdated. I just recently self-compiled it for another project. This broke the test as it does each request for a block with another token, the old (but outdated) version always uses the same token.

What do you think should be done now? Do you have the current version of libcoap coap-client in the test environment aswell? I will research on why this was done.

I found another issue too which seemed to also be happening in the tests. I will merge it's fix when we decided on what to do with the other issue

PR #629, opened 2025-10-20
FIX: response source address on [::]-bound servers are wrong

Theoyor:
When there are multiple addresses (a,b) on an interface and the coap-server is bound to [::] a client bound to a, sending to b received the response from a. This causes severe issues. 

Fixed in this PR handing over the control message to the response and inverting src and destination. I can not add a unit test for this case, as multiple addresses would need to be created in the test environment. 

@jkralik can you have a look?

## Summary by CodeRabbit

* **Refactor**
  * Centralized attachment and reuse of control information (interface index and local address) for incoming and outgoing UDP messages.
  * Server now tracks connections per remote/local address pair, improving per-address connection handling and response routing.

* **Chores**
  * Added IDE launch configuration for local development.
  * Expanded ignore rules to include IDE files.

### DIFF
--- a/udp/client/controlmessage_test.go
+++ b/udp/client/controlmessage_test.go
@@ -31,10 +31,10 @@ func TestSetControlInformationUnicastCopiesAddress(t *testing.T) {
 	stored := cc.localAddr.Load()
 	require.NotNil(t, stored)
 	expected := append(net.IP(nil), dst...)
-	require.True(t, (*stored).Equal(expected))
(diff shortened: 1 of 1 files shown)


Theoyor:
Hi @jkralik I tested again and it works well. Thanks for your additions and review. Would be lovely, if you merge it. Thank you!

jkralik:
@Theoyor Thank you for your contribution 😃

Issue #581, opened 2024-09-16
sendmsg invalid argument on udp connections with go-coap version >= 3.2.0


kgoguevvoi:
When attempting to upgrade our server which relies on go-coap v2 to v3 (latest) we encountered an issue with udp connections not being able to be written to from the server.  As an attempt to check if the latest code is working we resorted to using the simple coap example bundled with the repository.  When running the server and the client we are met with `sendmsg: invalid argument` on versions >= 3.2.0

The issue is not present on tag v3.1.6. However this presents issues for anyone that wants to use the latest version of the go-coap library.

The issue has been tested on the following platforms
```
* Linux 6.8.0-40-generic #40~22.04.3-Ubuntu SMP PREEMPT_DYNAMIC Tue Jul 30 17:30:19 UTC 2 x86_64 x86_64 x86_64 GNU/Linux
* Linux 6.10.9-arch1-2 #1 SMP PREEMPT_DYNAMIC Tue, 10 Sep 2024 14:37:32 +0000 x86_64 GNU/Linux
```

How to reproduce: 
1. Checkout the tag 3.2.0 and run the simple example server then client
2. Server fails to respond to client with invalid argument 
```
2024/09/16 15:31:27 ClientAddress 127.0.0.1:52765, Code: GET, Token: 0ab50dc5d0fdc417, Path: /a, Type: Confirmable, MessageID: 45080
udp: cannot write response: cannot write request: write udp [::]:5688->127.0.0.1:52765: sendmsg: invalid argument
2024/09/16 15:31:28 Error sending request: cannot write request: cannot write request: context deadline exceeded
exit status 1

```
3.  Checkout the tag 3.1.6 and run the simple example server then client
4.  Server responds as expected:
```
2024/09/16 15:36:41 ClientAddress 127.0.0.1:52530, Code: GET, Token: f29a0fd787cdc463, Path: /a, Type: Confirmable, MessageID: 53194
2024/09/16 15:36:41 Response payload: Code: Content, Token: f29a0fd787cdc463, ContentFormat: text/plain; charset=utf-8, Type: Acknowledgement, MessageID: 53194, PayloadLen: 11
```

Expected:
No error on server side with go-coap versions > 3.1.6

jlevesy:
From our investigation, It appears that this comes from https://github.com/plgd-dev/go-coap/pull/510.
[...]

PR #641, opened 2026-01-20
Fix sending to IPv4 addresses from an IPv6 dual-stack socket (#640)

### LINKED ISSUE #640: sendmsg: invalid argument on macOS via IPv4
Hello!

I am trying to update our codebase from go-coap v2 to v3. However, I encountered an issue where CoAP responses to calls received via IPv4 fail on macOS with an error similar to the following:

```
udp: cannot write response: cannot write request: write udp [::]:5688->127.0.0.1:58778: sendmsg: invalid argument
```

The issue might be related to a similar, already fixed issue on Linux: https://github.com/plgd-dev/go-coap/issues/581.

My platform is:

```
╰─$ uname -a
Darwin ****** 25.2.0 Darwin Kernel Version 25.2.0: Tue Nov 18 21:09:56 PST 2025; root:xnu-12377.61.12~1/RELEASE_ARM64_T6041 arm64
```

To reproduce, the following code can be used:

```go
package main

import (
	"bytes"
	"log"

	"github.com/plgd-dev/go-coap/v3"
)

func handleTest(w mux.ResponseWriter, r *mux.Message) {
	err := w.SetResponse(codes.Content, message.TextPlain, bytes.NewReader([]byte("hello world")))
	if err != nil {
		log.Println(err)
	}
}

func main() {
	r := mux.NewRouter()
	r.Handle("/test", mux.HandlerFunc(handleTest))
	log.Fatal(coap.ListenAndServe("udp", ":5688", r))
}
```

Run with e.g. `go run main.go` on macOS.

Then, submit the CoAP call via IPv4, e.g. use `coap-client coap://127.0.0.1:5688/test` from [libcoap](https://libcoap.net/). The server should print out the error message and the client fails after a while.

Calling via IPv6, e.g. `coap-client coap://\[::1\]:5688/test` or `coap-client coap://localhost:5688/test` works flawlessly.

I would expect the server to answer via IPv4 and IPv6.

The issue does not occur when using DTLS via pion/dtls.

Does someone have an idea?

ltlobaro:
This PR resolves sending packets to IPv4 adresses on an IPv6 dual-stack socket. It resolves #640 and is probably related to #581.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved UDP connection handling for IPv4/IPv6 compatibility, enabling better message transmission between different protocol versions.

* **Tests**
  * Added test coverage for sending IPv4 traffic from IPv6 sockets to ensure cross-protocol compatibility.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/net/connUDP_internal_test.go
+++ b/net/connUDP_internal_test.go
@@ -21,14 +21,9 @@ const (
 
 func TestUDPConnWriteWithContext(t *testing.T) {
 	iface := getInterfaceIndex(t)
-
 	ifaceIpv4 := getIfaceAddr(t, iface, true)
 	peerAddr := ifaceIpv4.String() + ":2154"
 	b, err := net.ResolveUDPAddr(udpNetwork, peerAddr)
-	require.NoError(t, err)
-
-	ifaceIpv6 := getIfaceAddr(t, iface, false)
-	v6, err := net.ResolveUDPAddr(udpNetwork, "["+ifaceIpv6.String()+"]:2154")
 	require.NoError(t, err)
 
 	ctxCanceled, ctxCancel := context.WithCancel(context.Background())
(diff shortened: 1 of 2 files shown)


jkralik:
Thx for contributing. Good job 👍

PR #642, opened 2026-01-22
fix: set UDP source address to original destination IP for Fly.io

denysvitali:
When using UDP on Fly.io, packets arrive at a public IP but the socket is bound to an internal private address. This commit ensures that responses are sent from the same IP that the client sent to, which is critical for correct routing.

The fix:
1. Store the original destination IP (from ControlMessage.Dst) in the session
2. Set ControlMessage.Src to that IP when sending responses

This ensures that when a client sends to Fly.io's public IP, the response is sent from that same IP, not from the internal private address.

See [this report](https://community.fly.io/t/golang-coap-server-over-udp-doesnt-work-tcp-does/20062/4) as well by @lyubchev.

I was able to get this working after using this patch.

## Summary by CodeRabbit

* **Bug Fixes**
  * Keepalive pings now consistently use the configured network interface and source address.
  * Responses preserve the original destination IP, improving message routing accuracy.
  * Existing message control data is preserved when sending responses.

* **Tests**
  * Added coverage for received-message responses, message writes, and keepalive pings.
  * Added validation for interface metadata and behavior when control information is unavailable.

### DIFF
--- a/udp/client/conn.go
+++ b/udp/client/conn.go
@@ -602,6 +602,9 @@ func (cc *Conn) AsyncPing(receivedPong func()) (func(), error) {
 	req.SetCode(codes.Empty)
 	mid := cc.GetMessageID()
 	req.SetMessageID(mid)
+	// Keepalive pings must reuse the original destination IP as the packet source
+	// (e.g. Fly.io public IP), same as regular request/response writes.
+	cc.upsertControlInformation(req)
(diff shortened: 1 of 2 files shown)


jkralik (commented):
Pls, fix build and tests.

denysvitali:
Sorry for getting back to you so late. Re-adapted the PR (and rebased). Wdyt?

Issue #644, opened 2026-02-25
piondtls.Config deprecation


sandrolain:
Hi,
`staticcheck` points out that the use of pion/dtls "Config" as a DTLS configuration has been deprecated in favor of the builder with options (I assume it will be removed with the release of DTLS 1.3).

> SA1019: piondtls.Config is deprecated: prefer the options-based APIs (`*WithOptions`) to construct immutable configs, This will be removed in the next major version. (staticcheck)

Is an integration or migration planned?
Possibly, I could help integrate builders similar to the existing ones with "WithOptions."

> NewDTLSListener => NewDTLSListenerWithOptions

Let me know!
Thanks for your work!

jkralik:
I liked it, so I merged it and also released the new v3.5.0. Thank you! 🙂

PR #645, opened 2026-02-26
Integrate DTLS (pion/dtls) configuration using options instead of deprecated *dtls.Config{}

sandrolain:
Hello, as mentioned in issue #644, with this PR I propose adapting the client and server methods to use the new options configuration provided by `pion/dtls`, which is now deprecating `*Config{}`.

First, I had to update the version of `pion/dtls` to the latest release, which introduced the deprecation of `*Config{}`, and consequently also the minimum go version to 1.24.

The solution I would like to propose introduces new structures to represent `pion/dtls` options, providing an atomic and abstract representation of these options while also avoiding confusion with some existing methods that use options for coap functionality. This structure would also allow for the complete abstraction of `pion/dtls` in the future.

```go
import (
  piondtls "github.com/pion/dtls/v3"
)

type DTLSClientOptions struct {
  opts []piondtls.ClientOption
}

func NewDTLSClientOptions(opts ...piondtls.ClientOption) DTLSClientOptions {
  return DTLSClientOptions{opts: opts}
}
```

Since some "WithOptions" methods already existed, I preferred to adapt the existing ones without renaming them, rather than adding duplicates, using generics to accept both the deprecated `*Config{}` and the new options structure as configuration parameters.
This will allow, at a later date, when deprecated methods and structures are removed from `pion/dtls` and the decision is made to update the dependency, the generic signature can be removed and only the new method will be retained.

```go
import (
  piondtls "github.com/pion/dtls/v3"
)

dtlsOpts := dtls.NewDTLSClientOptions(
  piondtls.WithCertificates(*certificate),
  piondtls.WithExtendedMasterSecret(piondtls.RequireExtendedMasterSecret),
  piondtls.WithRootCAs(certPool),
  piondtls.WithInsecureSkipVerify(false),
)
co, err := dtls.Dial("localhost:5688", dtlsOpts)
```

I also updated the examples to reflect the new method and integrated the tests and the readme.

_Working on Mac I added a fix to a test (TestTLSListenerCheckForInfinitLoop) that was failing due to compatibility issues._

## Summary by CodeRabbit

* **New Features**
  * Options-based DTLS API for client and server configuration with backwards-compatible support for legacy configs

* **Documentation**
  * README updated with migration guidance, deprecation notices, rationale, and options-based examples/references

* **Examples**
  * Multiple DTLS examples converted/added to show PSK and PKI usage via the new options API

* **Tests**
  * New tests covering the options-based DTLS client/server flows

* **Chores**
  * Updated Go toolchain requirement and bumped DTLS-related dependencies


jkralik:
Thank you for patient and contribution. Good job :)

Issue #603, opened 2025-04-01
"CoAP ping" is acknowleged


svenschwermer:
`IsPing` correctly identifying empty confirmable messages over unreliable transport as "CoAP pings" (RFC 7252).
https://github.com/plgd-dev/go-coap/blob/71407dadc3b4066060ee1e32095311160b6d058c/message/message.go#L53-L58

However, the response is an acknowledgment which is incorrect. It should be a Reset Message (RST).
https://github.com/plgd-dev/go-coap/blob/71407dadc3b4066060ee1e32095311160b6d058c/udp/client/conn.go#L645-L658

Quoting RFC 7252:
> Provoking a Reset message (e.g., by sending an Empty Confirmable message) is also useful as an inexpensive check of the liveness of an endpoint ("CoAP ping").

and

> "*" means that the combination is not used in normal operation but only to elicit a Reset message ("CoAP ping").

where `*` refers to an empty confirmable message.

PR #655, opened 2026-04-08
fix: CoAP Ping response uses RST instead of ACK per RFC 7252 §4.2

### LINKED ISSUE #486: Server response with ACK to Empty CON (CoAP Ping
Due to [rfc7252](https://datatracker.ietf.org/doc/html/rfc7252):

> the four message types can be used as in Table 1.  "*" means that the combination is not used in normal operation but only to elicit a Reset message ("CoAP ping").

In clientconn.go I can observe that also an ACK is accepted as Ping response:
https://github.com/plgd-dev/go-coap/blob/ea0ff53e834887b79cc696bb06914a636bd1b6e4/udp/client/conn.go#L474

We did observe, that the server indeed sends back an ACK and not a Reset message when sending a CoAP Ping.

sandrolain:
Fixes #486

**Files changed:** `udp/client/conn.go` (+3), `udp/client/conn_test.go` (+54)

**Problem:** `sendPong()` was setting `message.Acknowledgement` on the response
to a Confirmable empty message (CoAP Ping). RFC 7252 §4.2 requires the receiver
to reply with a matching Reset message, not an Acknowledgement. This caused
interoperability failures with compliant CoAP clients (reported in #603, #486).

**Fix:** One-line change — `message.Acknowledgement` → `message.Reset` in
`udp/client/conn.go`. A new test `TestConnPingResponseIsReset` verifies the
correct message type is sent.

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated the UDP client’s handling of confirmable ping requests to return a Reset response while preserving the original message ID.

* **Tests**
  * Added coverage to verify ping responses are Reset messages with the correct response code and matching message ID.

### DIFF
--- a/udp/client/conn.go
+++ b/udp/client/conn.go
@@ -652,7 +652,9 @@ func (cc *Conn) sendPong(w *responsewriter.ResponseWriter[*Conn], r *pool.Messag
 		cc.errors(fmt.Errorf("cannot send pong response: %w", err))
 	}
 	if r.Type() == message.Confirmable {
-		w.Message().SetType(message.Acknowledgement)
+		// RFC 7252 §4.2: An Empty Confirmable message (CoAP Ping) is rejected
+		// with a matching Reset message.
+		w.Message().SetType(message.Reset)
 		w.Message().SetMessageID(r.MessageID())
 	} else {
 		if w.Message().Type() != message.Reset {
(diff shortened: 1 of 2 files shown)


jkralik (approved):
Thank you for your contribution 👍

PR #660, opened 2026-06-16
659 bug blockwise windows udp

### LINKED ISSUE #659: Bug: Blockwise responses loop forever on Windows — `WriteMsgUDP` returns `n=0` but delivers the datagram, treated as `ErrWriteInterrupted`
### Environment
- go-coap: `v3.5.1`
- Go: `go1.26.4`
- OS: Windows (amd64)
- Transport: UDP

### Summary
On Windows, every server response triggers the error
`only part data was written to socket` (`ErrWriteInterrupted`), because
`(*net.UDPConn).WriteMsgUDP` reports `n=0` **even though the datagram is delivered
correctly**. For single-datagram responses this is harmless (the bogus error is only
logged and the client still gets its answer). For **blockwise responses (> 1 block)**
it breaks the multi-block state machine and the client/server loop **forever** while
allocating memory.

### Symptom
```
udp: cannot write response: cannot write request: only part data was written to socket
```
…repeated endlessly, with the request handler being re-invoked for every block request.

### Root cause
`(*net.UDPConn).WriteMsgUDP(b, oob, addr)` returns `n=0, err=nil` on Windows while
the datagram is actually sent. go-coap's UDP write path treats that as a partial write:

- `net/connUDP.go` — `writeTo()` uses the raw `c.connection.WriteMsgUDP(buffer, cmb, raddr)`
  for an unconnected server listener (`supportsOverrideRemoteAddr == true`), instead of
  the `ipv4.PacketConn.WriteTo` wrapper that works on Windows.
- `net/connUDP.go` — `writeWithCfg()` then does `if n != len(buffer) { return ErrWriteInterrupted }`,
  so on Windows it **always** returns `ErrWriteInterrupted`.
- `net/error.go` — `ErrWriteInterrupted = errors.New("only part data was written to socket")`.

#### Why small responses pass but blockwise loops
- **Single datagram:** datagram is delivered, error is only logged → request succeeds.
- **Blockwise (Block2):** block 0 is delivered, but the server sees an error, aborts the
  sending-message session / closes the connection. The client requests the next block →
  the server handles it as a brand-new request → re-runs the handler → re-sends block 0 →
  errors again → **infinite loop**.
- Block size is irrelevant (reproduced with SZX256 / SZX512 / SZX1024) → this is **not** an
  MTU/size problem, it's the bogus `n=0` return value.

### Minimal reproduction (stdlib only, isolates the root behavior)
Run on Windows:

```go
package main

import (
	"fmt"
	"net"
	"time"

	"golang.org/x/net/ipv4"
)

func main() {
	// Unconnected UDP server listener (same as coapNet.NewListenUDP).
	srv, _ := net.ListenUDP("udp4", &net.UDPAddr{IP: net.IPv4(127, 0, 0, 1)})
	defer srv.Close()
	p := ipv4.NewPacketConn(srv)
	_ = p.SetControlMessage(ipv4.FlagDst|ipv4.FlagInterface|ipv4.FlagSrc, true)

	srvAddr := srv.LocalAddr().(*net.UDPAddr)
	cli, _ := net.DialUDP("udp4", nil, srvAddr)
	defer cli.Close()
	_, _ = cli.Write([]byte("hi"))

	buf := make([]byte, 2048)
	_, _, _, raddr, _ := srv.ReadMsgUDP(buf, make([]byte, 1024))

	// Server replies. WriteMsgUDP is what go-coap's writeTo() uses.
	_ = cli.SetReadDeadline(time.Now().Add(500 * time.Millisecond))
	marker := []byte("DELIVERED-DESPITE-N0")
	n, _, err := srv.WriteMsgUDP(marker, nil, raddr)
	rn, _ := cli.Read(make([]byte, 64))
	fmt.Printf("WriteMsgUDP: n=%d err=%v ; client received %d bytes\n", n, err, rn)

	// The working alternative wrapper:
	for _, sz := range []int{300, 1024, 1100, 2000} {
		pl := make([]byte, sz)
		wm, _, _ := srv.WriteMsgUDP(pl, nil, raddr)
		wp, _ := p.WriteTo(pl, &ipv4.ControlMessage{Src: net.IPv4(127, 0, 0, 1)}, raddr)
		fmt.Printf("size=%4d  WriteMsgUDP n=%-4d  ipv4.PacketConn.WriteTo n=%-4d\n", sz, wm, wp)
	}
}
```

Output on Windows:
```
WriteMsgUDP: n=0 err=<nil> ; client received 20 bytes
size= 300  WriteMsgUDP n=0     ipv4.PacketConn.WriteTo n=300
size=1024  WriteMsgUDP n=0     ipv4.PacketConn.WriteTo n=1024
size=1100  WriteMsgUDP n=0     ipv4.PacketConn.WriteTo n=1100
size=2000  WriteMsgUDP n=0     ipv4.PacketConn.WriteTo n=2000
```

`WriteMsgUDP` returns `n=0` for every size (yet delivers the data); `ipv4.PacketConn.WriteTo`
returns the correct `n`.

### Expected behavior
A successfully sent server response (including each blockwise block) should not produce
`ErrWriteInterrupted`, and a large blockwise response should complete instead of looping.

### Actual behavior
Every UDP server response returns `ErrWriteInterrupted` on Windows; blockwise responses
loop forever and leak memory.

### Suggested fix
In `net/connUDP.go` `writeTo()`, use the already-available `packetConn.WriteTo`
(`ipv4.PacketConn.WriteTo` / `ipv6.PacketConn.WriteTo`) for the unconnected listener path
instead of the raw `c.connection.WriteMsgUDP`, which is unreliable on Windows.
Alternatively, treat `n == 0 && err == nil` from `WriteMsgUDP` as success on Windows.

### Workaround (consumer side)
No `coapNet` option avoids the `WriteMsgUDP` path, so until this is fixed, affected
blockwise tests/usage are skipped on Windows; everything works on Linux.

niondir:
fixes #659 

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved UDP multicast write error aggregation and preserved consistent success/failure return behavior across interfaces/sources.
  * Refined short-write interruption handling to flag only genuine partial writes.
  * Updated portable UDP write behavior documentation and aligned Windows UDP write-reporting expectations to ensure full datagram counts.
* **Tests**
  * Added UDP regression tests covering IPv4/IPv6 delivery, dual-stack listeners, and unconnected server write behavior.
  * Added Windows-only checks for `WriteMsgUDP`-reported byte counts.
  * Added a UDP CoAP blockwise end-to-end regression test for large responses to prevent repeated handler invocation.

### DIFF
--- a/net/connUDP.go
+++ b/net/connUDP.go
@@ -517,24 +517,17 @@ func (c *UDPConn) writeTo(raddr *net.UDPAddr, cm *ControlMessage, buffer []byte)
 		return c.connection.Write(buffer)
 	}
 
-	var cmb []byte
-
-	if cm != nil && IsIPv6(raddr.IP) {
-		m := &ipv6.ControlMessage{
-			Src:     cm.Src,
-			IfIndex: cm.IfIndex,
-		}
-		cmb = m.Marshal()
-	} else if cm != nil {
-		m := &ipv4.ControlMessage{
-			Src:     cm.Src,
-			IfIndex: cm.IfIndex,
-		}
-		cmb = m.Marshal()
+	// For an unconnected listener we use the ipv4/ipv6 PacketConn.WriteTo wrapper
+	// instead of the raw (*net.UDPConn).WriteMsgUDP. On Windows WriteMsgUDP reports
+	// n=0 even though the datagram is delivered correctly, which made writeWithCfg
+	// treat every server response as a partial write (ErrWriteInterrupted) and broke
+	// blockwise transfers (endless request/response loop). PacketConn.WriteTo returns
+	// the correct number of written bytes on all platforms.
+	p, err := newPacketConnWithAddr(raddr, c.connection)
+	if err != nil {
+		return 0, err
 	}
-
-	i, _, err := c.connection.WriteMsgUDP(buffer, cmb, raddr)
-	return i, err
+	return p.WriteTo(buffer, cm, raddr)
 }
 
 type UDPWriteCfg struct {
--- a/net/connUDP.go
+++ b/net/connUDP.go
@@ -408,22 +408,22 @@ func (c *UDPConn) writeMulticastWithInterface(raddr *net.UDPAddr, buffer []byte,
 	if IsIPv6(raddr.IP) {
 		netType = "udp6"
 	}
-	var errors []error
+	var errs []error
 	for _, ip := range convAddrsToIps(filterAddressesByNetwork(netType, ifaceAddrs)) {
 		ipAddr := ip
 		opt.Source = &ipAddr
 		err = c.writeToAddr(opt.Iface, opt.Source, opt.HopLimit, raddr, buffer)
 		if err != nil {
-			errors = append(errors, err)
+			errs = append(errs, err)
 		}
 	}
-	if errors == nil {
+	if errs == nil {
 		return nil
 	}
-	if len(errors) == 1 {
-		return errors[0]
+	if len(errs) == 1 {
+		return errs[0]
 	}
-	return fmt.Errorf("%v", errors)
+	return fmt.Errorf("%v", errs)
 }
 
 func (c *UDPConn) writeMulticastToAllInterfaces(raddr *net.UDPAddr, buffer []byte, opt MulticastOptions) error {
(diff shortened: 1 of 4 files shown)


jkralik:
Hi @niondir, I created an MR that includes your changes along with a few additional improvements from my side: https://github.com/plgd-dev/go-coap/pull/661
Thanks for your contribution 😄

PR #677, opened 2026-07-07
fix(blockwise): return 4.08 on incomplete Block1 and make receive cache limit configurable

### LINKED ISSUE #664: RFC 7252 / 7959 Conformance Issues — Malformed-Datagram Session Teardown and Block-wise Transfer Gaps
Hi maintainers,
While reviewing `go-coap`'s UDP/NoSec server and block-wise transfer implementation against RFC 7252 and RFC 7959, we noticed several places where the current behavior appears to differ from the specification. Each item below cites the relevant RFC text alongside the corresponding source code, and includes a minimal PoC in the accompanying subdirectory. We hope this is helpful for improving conformance.

**PoC bundle** (attached):
- `CoAP_011/` — malformed-datagram session teardown (item 1).
- `blockwise_pocs/` — block-wise conformance issues (items 2–5), one binary with an `-issue` flag.

Both PoCs build against a go-coap checkout placed at `../go-coap-sut` (clone or symlink). See each subdirectory's `README.md` for details.

**Reviewed go-coap version**:
- Tag: `v3.5.1` (`v3.5.1-1-g8d7399b`)
- Commit: `8d7399baf06079dad773177a557740c935f6a563`

---

### 1. Malformed UDP Datagram Tears Down the Entire Per-Peer Connection

**RFC Reference**: RFC 7252 Section 4.2

> "A recipient MUST either (a) acknowledge a Confirmable message with an Acknowledgement message or (b) reject the message if the recipient lacks context to process the message properly, including situations where the message is Empty, uses a code with a reserved class (1, 6, or 7), or has a message format error."

**Analysis**:
In the UDP server read loop, any datagram that fails to decode causes the server to close the whole per-peer connection for that remote address. A 4-byte datagram with a zeroed first byte (CoAP version ≠ 1) is enough. Because the per-peer `*client.Conn` carries all server-side state for that remote endpoint (Observe subscriptions, in-flight block-wise transfer caches, token/MID dedup tables), a single malformed datagram silently destroys all of it. RFC 7252 §4.2 prescribes a per-message response (ACK or reject); it does not permit tearing down the recipient's entire session state as a side effect of one bad datagram.

**Source Code Evidence** (`udp/server/server.go`, `udp/client/conn.go`, `udp/coder/coder.go`):

```go
// udp/server/server.go:179-183 (Serve loop)
err = cc.Process(cm, buf)
if err != nil {
    s.closeConnection(cc)   // <-- drops ALL state for this peer
    s.cfg.Errors(fmt.Errorf("%v: cannot process packet: %w", cc.RemoteAddr(), err))
}
```

```go
// udp/client/conn.go:923-931
func (cc *Conn) Process(cm *coapNet.ControlMessage, datagram []byte) error {
    req := cc.AcquireMessage(cc.Context())
    defer cc.ReleaseMessage(req)
    _, err := req.UnmarshalWithDecoder(coder.DefaultCoder, datagram)
    if err != nil {
        return err              // -> bubbles up to Serve loop -> closeConnection
    }
    ...
}
```

```go
// udp/coder/coder.go:97-117
func (c *Coder) Decode(data []byte, m *message.Message) (int, error) {
    size := len(data)
    if size < 4 {
        return -1, ErrMessageTruncated
    }
    if data[0]>>6 != 1 {
        return -1, ErrMessageInvalidVersion     // <-- 4 zero bytes hit this
    }
    ...
    tokenLen := int(data[0] & 0xf)
    if tokenLen > 8 {
        return -1, message.ErrInvalidTokenLen
    }
    ...
}
```

The connection is keyed by the 4-tuple (`getConnKey`, `udp/server/server.go:278-292`); `closeConnection` (`udp/server/server.go:110`) closes the `*client.Conn` and removes it from the connection map, destroying all associated state.

**Verification**: Confirmed dynamically. A minimal in-process PoC starts a `go-coap` UDP Observe server, establishes an Observe subscription over a raw UDP socket, confirms 2.05 notifications flow, then injects a single 4-byte datagram `{0x00,0x00,0x00,0x00}` (version=0 → `ErrMessageInvalidVersion`) over the same socket. The server logs `cannot process packet: message has invalid version`, the notification goroutine stops with `context canceled`, and no further notifications arrive in a 3-second observation window (expected ~20 at 150ms cadence). The PoC exits with code 1 (violation confirmed). **Impact**: Denial of Service — a single 4-byte datagram kills all of a peer's server-side state (Observe subscriptions, block-wise transfers). In NoSec deployments the UDP source address is unauthenticated, so an off-path attacker who can guess the victim's source `ip:port` can spoof this 4-tuple and repeatedly disrupt the victim's sessions. **Mitigating factors**: (1) session-scoped, not whole-server; (2) DTLS deployments are immune (an attacker cannot inject decodable-as-this-peer datagrams); (3) TCP framing errors are legitimately connection-fatal (stream desync), so the teardown is defensible there. **Severity**: Medium (availability; trivial to trigger; off-path spoofable in NoSec). A minimal fix is to drop the bad datagram and continue serving rather than calling `closeConnection`:

```diff
  err = cc.Process(cm, buf)
  if err != nil {
-     s.closeConnection(cc)
-     s.cfg.Errors(fmt.Errorf("%v: cannot process packet: %w", cc.RemoteAddr(), err))
+     s.cfg.Errors(fmt.Errorf("%v: dropping malformed packet: %w", cc.RemoteAddr(), err))
  }
```

**PoC**: `CoAP_011/` — `go run .` (see `CoAP_011/README.md`). Exit code 1 = violation confirmed.

---

### 2. Missing Blocks Silently Dropped — No 4.08 (Request Entity Incomplete)

**RFC Reference**: RFC 7959 Section 2.5

> "If not all previous blocks are available at the server at the time of processing the final block, the transfer fails and error code 4.08 (Request Entity Incomplete, Section 2.9.2) MUST be returned."

**Analysis**:
During block-wise assembly of a request body (Block1), the server appends an incoming block's payload only when `off == payloadSize` (i.e., the block's expected offset matches the currently accumulated size). If a previous block is missing, `off != payloadSize` and the incoming block is silently dropped — no 4.08 is returned. The transfer then stalls until the `BlockwiseTransferTimeout` (3s default) expires and the cache entry is evicted. Additionally, the `receivingMessagesCache` is unbounded (no `MaxEntries`), so an attacker can open many concurrent partial transfers to consume memory.

**Source Code Evidence** (`net/blockwise/blockwise.go`):

```go
// net/blockwise/blockwise.go:807-828
off := num * szx.Size()
if off == payloadSize { //nolint:nestif
    payloadSize, err = copyToPayloadFromOffset(r, payloadFile, off)
    ...
    if !more {
        b.receivingMessagesCache.Delete(tokenStr)
        ...
        next(w, cachedReceivedMessage)
        return nil
    }
}
// if off != payloadSize: fall through, no 4.08, block silently dropped
```

```go
// net/blockwise/blockwise.go:171 — cache has no entry limit
receivingMessagesCache:    cache.NewCache[uint64, *messageGuard](),
sendingMessagesCache:      cache.NewCache[uint64, *pool.Message](),
```

The only 4.08 (`RequestEntityIncomplete`) emission in the blockwise package is at `blockwise.go:318`, on a different path (block1/block2 coexistence), not on the missing-previous-block path.

**Verification**: Confirmed dynamically. The PoC sends Block1 NUM=0 (M=1), then Block1 NUM=2 (M=0) without sending block NUM=1. The server returns 2.31 Continue (code `0x5F`) instead of 4.08, silently dropping the out-of-order block. **Impact**: A client that sends blocks out of order (or omits a block) causes the server to silently stall the transfer for 3 seconds rather than immediately returning 4.08. The unbounded cache permits memory exhaustion via many concurrent partial transfers. **Severity**: Low–Medium (robustness/DoS; 3s stall per malformed transfer; memory pressure with many concurrent transfers). **PoC**: `blockwise_pocs/` — `go run . -issue block_009`.

---

### 3. Content-Format Not Compared Across Blocks During Assembly

**RFC Reference**: RFC 7959 Section 2.3

> "If blocks of a request arrive at a server with mismatching Content-Format Options, the server MUST NOT assemble them into a single request; this usually leads to a 4.08 (Request Entity Incomplete, Section 2.9.2) error response on the mismatching block."

**Analysis**:
The block assembly function `getPayloadFromCachedReceivedMessage` compares the ETag option between the incoming block and the cached (previously received) message, but it does not compare the Content-Format option. A client that sends blocks with different Content-Format values will have them silently assembled into a single request body, which is then forwarded to the application handler with whatever Content-Format the first block carried.

**Source Code Evidence** (`net/blockwise/blockwise.go`):

```go
// net/blockwise/blockwise.go:628-656
func (b *BlockWise[C]) getPayloadFromCachedReceivedMessage(r, cachedReceivedMessage *pool.Message) (*memfile.File, int64, error) {
    payloadFile, ok := cachedReceivedMessage.Body().(*memfile.File)
    ...
    rETAG, errETAG := r.GetOptionBytes(message.ETag)
    cachedReceivedMessageETAG, errCachedReceivedMessageETAG := cachedReceivedMessage.GetOptionBytes(message.ETag)
    switch {
    case errETAG == nil && errCachedReceivedMessageETAG != nil:
        ...  // ETag presence/absence compared
    case errETAG != nil && errCachedReceivedMessageETAG == nil:
        ...
    case !bytes.Equal(rETAG, cachedReceivedMessageETAG):
        // ETag changed - drop data and set new ETAG
        ...
    }
    // NO Content-Format comparison anywhere in this function
    payloadSize, err := cachedReceivedMessage.BodySize()
    ...
    return payloadFile, payloadSize, nil
}
```

A grep for `ContentFormat` / `contentFormat` in `net/blockwise/blockwise.go` returns no hits within the assembly path.

**Verification**: Confirmed dynamically. The PoC sends Block1 NUM=0 with Content-Format `text/plain` (0), then Block1 NUM=1 with Content-Format `application/json` (50). The server assembles both blocks and returns 2.04 Changed (code `0x44`) instead of 4.08. **Impact**: A request body assembled from blocks with differing Content-Format options is delivered to the application handler with an inaccurate Content-Format, potentially causing misinterpretation of the payload. **Severity**: Low (correctness; requires a misbehaving or malicious client; no memory-safety impact). **PoC**: `blockwise_pocs/` — `go run . -issue block_006`.

---

### 4. Duplicate Block Option Not Rejected

**RFC Reference**: RFC 7959 Section 2.1

> "When it is present in a CoAP message, it MUST be processed (or the message rejected); therefore, it is identified as a Critical option. Either Block option MUST NOT occur more than once in a single message."

**Analysis**:
The option decoder does not enforce the "MUST NOT occur more than once" constraint for non-repeatable options. `GetOptionUint32` uses `Options.Find(id)`, which returns a `(firstIdx, lastIdx)` range, but the caller only reads `options[firstIdx]` — the first occurrence wins and subsequent duplicates are silently ignored. The Block1/Block2 option definitions do not set the Repeatable flag (`message/option.go:152-153`), but this flag is not consulted during decoding to reject duplicates.

**Source Code Evidence** (`message/options.go`, `message/option.go`):

```go
// message/options.go:200-208
func (options Options) GetUint32(id OptionID) (uint32, error) {
    firstIdx, _, err := options.Find(id)   // <-- ignores lastIdx; duplicates silently accepted
    if err != nil {
        return 0, err
    }
    val, _, err := DecodeUint32(options[firstIdx].Value)
    return val, err
}
```

```go
// message/option.go:152-153 — Block1/Block2 not marked Repeatable, but
// nothing in the decode path rejects repeated occurrences.
Block2:        {ValueFormat: ValueUint, MinLen: 0, MaxLen: 3},
Block1:        {ValueFormat: ValueUint, MinLen: 0, MaxLen: 3},
```

The blockwise handler reads the Block option via `r.GetOptionUint32(message.Block1/Block2)` (`net/blockwise/blockwise.go:398, 404, 410`), which returns the first occurrence without checking for duplicates.

**Verification**: Confirmed dynamically. The PoC sends a GET with two identical Block2 options. The server processes the request using the first occurrence and returns 2.05 Content (code `0x45`) instead of rejecting with 4.00. **Impact**: A message with two Block2 options (e.g., one requesting block 0 and one requesting block 999) is processed using only the first, with no error. A malicious peer could use this to confuse block-wise state. **Severity**: Low (conformance; no memory-safety impact; the first-option-wins behavior is deterministic). **PoC**: `blockwise_pocs/` — `go run . -issue block_001`.

---

### 5. If-None-Match Not Restricted to Block1 NUM=0

**RFC Reference**: RFC 7959 Section 2.10

> "If-None-Match only works correctly on Block1 requests with (NUM=0) and MUST NOT be used on Block1 requests with NUM != 0."

**Analysis**:
The block-wise handler does not inspect the If-None-Match option at all. It is passed through to the application handler regardless of the Block1 NUM value. A PUT with Block1 NUM=5 and If-None-Match set will reach the application handler with If-None-Match present, where the application may incorrectly apply the "create only if not exists" semantics to a non-initial block.

**Source Code Evidence** (`net/blockwise/blockwise.go`):

A grep for `IfNoneMatch` / `IfNone` / `ifNoneMatch` in `net/blockwise/blockwise.go` returns no hits. The block1 processing path (`handleReceivedMessage` → `processReceivedMessage` at lines 397-408) reads Block1 and Size1 but never checks for the presence of If-None-Match or its interaction with the Block1 NUM.

**Verification**: Confirmed dynamically. The PoC sends Block1 NUM=0 (no If-None-Match), then Block1 NUM=1 (M=0) carrying an If-None-Match option. The server assembles the body and returns 2.04 Changed (code `0x44`) instead of rejecting with 4.00. **Impact**: The application handler may apply If-None-Match semantics to non-initial blocks, producing incorrect create-vs-replace behavior. **Severity**: Low (correctness; requires a misbehaving client; the application handler can independently guard against this). **PoC**: `blockwise_pocs/` — `go run . -issue block_007`.

---

[POC.zip](https://github.com/user-attachments/files/29170109/POC.zip)

jkralik:
Fixes #664 point 2

## Summary by CodeRabbit

* **New Features**
  * Added an option to cap in-flight incoming block-wise assemblies (default **1024**; set to `0` for unlimited) across UDP, TCP, and DTLS clients/servers.
  * Exposed a configurable setting for the receiving-messages cache limit.

* **Bug Fixes**
  * Strengthened receive-side validation for block-wise POST transfers with missing previous blocks or invalid final blocks.
  * When the receiving cache is full, now returns **ServiceUnavailable**; other incomplete transfers map to **RequestEntityIncomplete**.

* **Tests**
  * Added unit tests and an end-to-end regression test covering missing previous blocks, invalid final blocks, and bounded cache behavior.

### DIFF
--- a/dtls/client.go
+++ b/dtls/client.go
@@ -97,7 +97,7 @@ func Client(conn *dtls.Conn, opts ...udp.Option) *udpClient.Conn {
 	if cfg.BlockwiseEnable {
 		createBlockWise = func(cc *udpClient.Conn) *blockwise.BlockWise[*udpClient.Conn] {
 			v := cc
-			return blockwise.New(
+			bw := blockwise.New(
 				v,
 				cfg.BlockwiseTransferTimeout,
 				cfg.Errors,
@@ -105,6 +105,8 @@ func Client(conn *dtls.Conn, opts ...udp.Option) *udpClient.Conn {
 					return v.GetObservationRequest(token)
 				},
 			)
+			bw.SetReceivingMessagesCacheMaxEntries(cfg.BlockwiseReceivingMessagesCacheMaxEntries)
+			return bw
 		}
 	}
 
--- a/dtls/server/server.go
+++ b/dtls/server/server.go
@@ -218,7 +218,7 @@ func (s *Server) createConn(connection *coapNet.Conn, inactivityMonitor udpClien
 	if s.cfg.BlockwiseEnable {
 		createBlockWise = func(cc *udpClient.Conn) *blockwise.BlockWise[*udpClient.Conn] {
 			v := cc
-			return blockwise.New(
+			bw := blockwise.New(
 				v,
 				s.cfg.BlockwiseTransferTimeout,
 				s.cfg.Errors,
@@ -226,6 +226,8 @@ func (s *Server) createConn(connection *coapNet.Conn, inactivityMonitor udpClien
 					return v.GetObservationRequest(token)
 				},
 			)
+			bw.SetReceivingMessagesCacheMaxEntries(s.cfg.BlockwiseReceivingMessagesCacheMaxEntries)
+			return bw
 		}
 	}
 	session := NewSession(
--- a/net/blockwise/blockwise.go
+++ b/net/blockwise/blockwise.go
@@ -44,6 +44,8 @@ const (
 	moreBlocksFollowingMask = 0x8
 	// szxMask last 3bits represents SZX (SZX)
 	szxMask = 0x7
+	// max number of concurrent in-flight incoming block-wise assemblies.
+	DefaultReceivingMessagesCacheMaxEntries = 1024
 )
 
 // SZX enum representation for the size of the block: https://tools.ietf.org/html/rfc7959#section-2.2
@@ -135,18 +137,21 @@ type Client interface {
 }
 
 type BlockWise[C Client] struct {
-	cc                        C
-	receivingMessagesCache    *cache.Cache[uint64, *messageGuard]
-	sendingMessagesCache      *cache.Cache[uint64, *pool.Message]
-	errors                    func(error)
-	getSentRequestFromOutside func(token message.Token) (*pool.Message, bool)
-	expiration                time.Duration
+	cc                               C
+	receivingMessagesCache           *cache.Cache[uint64, *messageGuard]
+	receivingMessagesCacheMaxEntries int
+	sendingMessagesCache             *cache.Cache[uint64, *pool.Message]
+	errors                           func(error)
+	getSentRequestFromOutside        func(token message.Token) (*pool.Message, bool)
+	expiration                       time.Duration
 }
 
 type messageGuard struct {
 	*pool.Message
 	*semaphore.Weighted
 }
+
+var errReceivingMessagesCacheFull = errors.New("receiving messages cache is full")
 
 func newRequestGuard(request *pool.Message) *messageGuard {
 	return &messageGuard{
@@ -167,13 +172,23 @@ func New[C Client](
 		getSentRequestFromOutside = func(message.Token) (*pool.Message, bool) { return nil, false }
 	}
 	return &BlockWise[C]{
-		cc:                        cc,
-		receivingMessagesCache:    cache.NewCache[uint64, *messageGuard](),
-		sendingMessagesCache:      cache.NewCache[uint64, *pool.Message](),
-		errors:                    errors,
-		getSentRequestFromOutside: getSentRequestFromOutside,
-		expiration:                expiration,
+		cc:                               cc,
+		receivingMessagesCache:           cache.NewCache[uint64, *messageGuard](),
+		receivingMessagesCacheMaxEntries: DefaultReceivingMessagesCacheMaxEntries,
+		sendingMessagesCache:             cache.NewCache[uint64, *pool.Message](),
+		errors:                           errors,
+		getSentRequestFromOutside:        getSentRequestFromOutside,
+		expiration:                       expiration,
 	}
+}
+
+// SetReceivingMessagesCacheMaxEntries sets max entries for in-flight incoming
+// block-wise assemblies. Value 0 means unlimited.
+func (b *BlockWise[C]) SetReceivingMessagesCacheMaxEntries(maxEntries int) {
+	if maxEntries < 0 {
+		maxEntries = 0
+	}
+	b.receivingMessagesCacheMaxEntries = maxEntries
 }
 
 func bufferSize(szx SZX, maxMessageSize uint32) int64 {
@@ -321,6 +336,14 @@ func (b *BlockWise[C]) sendEntityIncomplete(w *responsewriter.ResponseWriter[C],
 	w.SetMessage(sendMessage)
 }
 
+func (b *BlockWise[C]) sendServiceUnavailable(w *responsewriter.ResponseWriter[C], token message.Token) {
+	sendMessage := b.cc.AcquireMessage(w.Message().Context())
+	sendMessage.SetCode(codes.ServiceUnavailable)
+	sendMessage.SetToken(token)
+	sendMessage.SetType(message.NonConfirmable)
+	w.SetMessage(sendMessage)
+}
+
 func wantsToBeReceived(r *pool.Message) bool {
 	hasBlock1 := r.HasOption(message.Block1)
 	hasBlock2 := r.HasOption(message.Block2)
@@ -346,30 +369,51 @@ func (b *BlockWise[C]) getSendingMessageCode(token uint64) (codes.Code, bool) {
 	return v.Data().Code(), true
 }
 
+type receiveRequestCall[C Client] struct {
+	w    *responsewriter.ResponseWriter[C]
+	r    *pool.Message
+	next func(w *responsewriter.ResponseWriter[C], r *pool.Message)
+}
+
+func (c receiveRequestCall[C]) forward() {
+	c.next(c.w, c.r)
+}
+
+func (b *BlockWise[C]) handleReceivedMessageWithErrorResponse(call receiveRequestCall[C], maxSZX SZX, maxMessageSize uint32) {
+	err := b.handleReceivedMessage(call.w, call.r, maxSZX, maxMessageSize, call.next)
+	if err == nil {
+		return
+	}
+	token := call.r.Token()
+	if errors.Is(err, errReceivingMessagesCacheFull) {
+		b.sendServiceUnavailable(call.w, token)
+	} else {
+		b.sendEntityIncomplete(call.w, token)
+	}
+	b.errors(fmt.Errorf("handleReceivedMessage(%v): %w", call.r, err))
+}
+
 // Handle middleware which constructs COAP request from blockwise transfer and send COAP response via blockwise.
 func (b *BlockWise[C]) Handle(w *responsewriter.ResponseWriter[C], r *pool.Message, maxSZX SZX, maxMessageSize uint32, next func(w *responsewriter.ResponseWriter[C], r *pool.Message)) {
 	if maxSZX > SZXBERT {
 		panic("invalid maxSZX")
 	}
+	call := receiveRequestCall[C]{
+		w:    w,
+		r:    r,
+		next: next,
+	}
 	token := r.Token()
 
 	if len(token) == 0 {
-		err := b.handleReceivedMessage(w, r, maxSZX, maxMessageSize, next)
-		if err != nil {
-			b.sendEntityIncomplete(w, token)
-			b.errors(fmt.Errorf("handleReceivedMessage(%v): %w", r, err))
-		}
+		b.handleReceivedMessageWithErrorResponse(call, maxSZX, maxMessageSize)
 		return
 	}
 	tokenStr := token.Hash()
 
 	sendingMessageCode, sendingMessageExist := b.getSendingMessageCode(tokenStr)
 	if !sendingMessageExist || wantsToBeReceived(r) {
-		err := b.handleReceivedMessage(w, r, maxSZX, maxMessageSize, next)
-		if err != nil {
-			b.sendEntityIncomplete(w, token)
-			b.errors(fmt.Errorf("handleReceivedMessage(%v): %w", r, err))
-		}
+		b.handleReceivedMessageWithErrorResponse(call, maxSZX, maxMessageSize)
 		return
 	}
 	more, err := b.continueSendingMessage(w, r, maxSZX, maxMessageSize, sendingMessageCode)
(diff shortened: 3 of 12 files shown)


palsivertsen:
The following finding is based on be2f7aeaf12833ca1f12fc4ca1d126df6d76419d

This fix does not correctly handle cases where the first block is missing. I tried making a block1 request, staring with the second block. Since I skipped the initial block I expected the server to immediately recognize that there's a missing block, but instead it happily returned the `2.31` status code. It's also worth noting that the request never reached my handler.


<details>
<summary>Here's a test file demonstrating the issue</summary>

```go
// Package coaprepro is a minimal, standalone reproduction of a go-coap
// block-wise (RFC 7959) server bug: a Block1 transfer whose very first
// message claims block NUM=1 (block 0 was never sent) is answered with
// 2.31 Continue instead of being rejected with 4.08 Request Entity
// Incomplete, per RFC 7959 section 2.9.2.
package coaprepro_test

import (
	"encoding/binary"
	"fmt"
	"net"
	"testing"
	"time"

	"github.com/plgd-dev/go-coap/v3/options"
	"github.com/plgd-dev/go-coap/v3/udp/server"
	"github.com/stretchr/testify/require"

	coapnet "github.com/plgd-dev/go-coap/v3/net"
	coapblockwise "github.com/plgd-dev/go-coap/v3/net/blockwise"
)

// CoAP header/option constants (RFC 7252) and Block1 encoding (RFC 7959),
// just enough to build one request and read one response code byte.
const (
	coapVersion     = 1
	typeConfirmable = 0
	codePUT         = 0x03

	codeRequestEntityIncomplete = 0x88 // 4.08 Request Entity Incomplete

	optionURIPath = 11
	optionBlock1  = 27

	payloadMarker = 0xFF

	blockSZX128 = 3 // block size = 2^(4+3) = 128 bytes
	blockSize   = 128
)

// codeStr renders a CoAP response code the way RFC 7252 documents it (e.g.
// "2.31", "4.08"), so test failures are easy to read.
func codeStr(code byte) string {
	return fmt.Sprintf("%d.%02d", code>>5, code&0x1f)
}

// encodeOptionLength encodes a delta or length value into its 4-bit nibble
// plus any extended bytes (RFC 7252 section 3.1).
func encodeOptionLength(length uint16) (byte, []byte) {
	switch {
	case length < 13:
		return byte(length), nil
	case length < 269:
		return 13, []byte{byte(length - 13)}
	default:
		extended := length - 269
		return 14, []byte{byte(extended >> 8), byte(extended)} //nolint:gosec
	}
}

// appendOption appends one option (as a delta from the previous option
// number) to buf, per RFC 7252 section 3.1.
func appendOption(buf []byte, delta uint16, value []byte) []byte {
	deltaNibble, deltaExt := encodeOptionLength(delta)
	lengthNibble, lengthExt := encodeOptionLength(uint16(len(value))) //nolint:gosec

	buf = append(buf, (deltaNibble<<4)|lengthNibble)
	buf = append(buf, deltaExt...)
	buf = append(buf, lengthExt...)
	buf = append(buf, value...)
	return buf
}

// encodeBlock1 encodes a Block1 option value (RFC 7959 section 2.1): NUM in
// the most significant bits, then the M (more) flag, then SZX.
func encodeBlock1(num uint32, more bool, szx byte) []byte {
	fieldValue := (num << 4) | uint32(szx&0x7)
	if more {
		fieldValue |= 0x8
	}
	switch {
	case fieldValue == 0:
		return []byte{0}
	case fieldValue < 1<<8:
		return []byte{byte(fieldValue)}
	default:
		return []byte{byte(fieldValue >> 8), byte(fieldValue)} //nolint:gosec
	}
}

// buildBlock1PUT builds a confirmable PUT to Uri-Path "x" carrying a single
// Block1 chunk (RFC 7959), ready to write to a UDP socket.
func buildBlock1PUT(messageID uint16, token []byte, num uint32, more bool, chunk []byte) []byte {
	buf := []byte{
		(coapVersion << 6) | (typeConfirmable << 4) | byte(len(token)),
		codePUT,
	}
	buf = binary.BigEndian.AppendUint16(buf, messageID)
	buf = append(buf, token...)

	// Options must be in ascending option-number order: Uri-Path (11)
	// before Block1 (27).
	buf = appendOption(buf, optionURIPath, []byte("x"))
	buf = appendOption(buf, optionBlock1-optionURIPath, encodeBlock1(num, more, blockSZX128))

	buf = append(buf, payloadMarker)
	buf = append(buf, chunk...)
	return buf
}

// newBlockwiseServer starts a real CoAP-over-UDP server on an ephemeral
// port with block-wise transfers enabled, routing every request to a
// handler that replies 2.04 Changed once a request is fully reassembled.
func newBlockwiseServer(t *testing.T) string {
	t.Helper()

	listener, err := coapnet.NewListenUDP("udp", "")
	require.NoError(t, err, "create listener")
	t.Cleanup(func() { _ = listener.Close() })

	router := mux.NewRouter()
	router.HandleFunc("/x", func(w mux.ResponseWriter, r *mux.Message) {
		if _, err := r.ReadBody(); err != nil {
			return
		}
		_ = w.SetResponse(codes.Changed, message.TextPlain, nil)
	})

	srv := server.New(
		options.WithMux(router),
		options.WithBlockwise(true, coapblockwise.SZX128, time.Minute),
	)
	go srv.Serve(listener) //nolint:errcheck
	t.Cleanup(srv.Stop)

	return listener.LocalAddr().String()
}

// TestFirstBlockNotZeroReturnsIncomplete asserts RFC 7959 section 2.9.2: a
// transfer whose first Block1 message is NUM=1 (block 0 was never sent)
// must be rejected with 4.08 Request Entity Incomplete, not treated as if
// NUM=1 were the start of the payload.
//
// Observed: the server replies 2.31 Continue (0x5f) instead.
func TestFirstBlockNotZeroReturnsIncomplete(t *testing.T) {
	addr := newBlockwiseServer(t)

	raddr, err := net.ResolveUDPAddr("udp", addr)
	require.NoError(t, err, "resolve server address")
	conn, err := net.DialUDP("udp", nil, raddr)
	require.NoError(t, err, "dial server")
	t.Cleanup(func() { _ = conn.Close() })

	// The very first message of this transfer claims to be block NUM=1;
	// block 0 was never sent.
	chunk := make([]byte, blockSize)
	req := buildBlock1PUT(1, []byte{0xDD}, 1, true, chunk)

	_, err = conn.Write(req)
	require.NoError(t, err, "write request")

	require.NoError(t, conn.SetReadDeadline(time.Now().Add(2*time.Second)), "set read deadline")
	buf := make([]byte, 2048)
	readLen, err := conn.Read(buf)
	require.NoError(t, err, "read response (server did not reply in time)")
	require.GreaterOrEqual(t, readLen, 2, "response too short to contain a CoAP code")

	responseCode := buf[1]
	require.Equal(t, byte(codeRequestEntityIncomplete), responseCode,
		"expected 4.08 Request Entity Incomplete when the first block isn't NUM=0, got %s",
		codeStr(responseCode))
}

```

</details>

